PR #8, opened 2022-05-22
minimum viable fuelup

binggh:
rough minimum viable fuelup that downloads latest core binaries to start building on Fuel Network and unpacks it into `$HOME/.fuelup`. Creating this PR up early for feedback and thoughts.

The following are installed through this script: 

The `forc-binaries` tarball from the sway repo:
`forc`, `forc-fmt`, `forc-explore`, `forc-explore`

And `fuel-core` from `fuel-core` repo

Discussion:

I tackled the rustup codebase and found it too big to use its concepts to make an MVP right now. Was also looking at foundry for inspiration and realized they were using just bash scripts to download and install `foundryup`, in the end decided to do it in a clap app even though it would probably be much faster if we just script everything, but would probably be harder to introduce features later.

I think maybe profiles is something we can move towards after we have some sort of an MVP ready, with other basic features.

Included the bin (built for x86_64-apple-darwin) as well for now

Immediate things to do for this PR:

- [ ] include a shell script named  `fuelup-init.sh`(as rustup/foundry does) to download this `fuelup` bin, setup the system (create a `/.fuelup` dir for example) and run `fuelup` automatically, so devs would only need to do `curl -L <url> | sh` to be able to set up forc binaries + fuel-core together.

After this PR:

- [ ] CI for `fuelup`, so we can release latest `fuelup` features/etc. as we iterate and provide more features for the toolchain.
- [ ] `fuelup default` to set the default for certain things (not sure how this should work - should we be able to set different versions for each of `forc`, `fuel-core`, etc?)

cc: @mitchmindtree @JoshuaBatty @adlerjohn for feedback


mitchmindtree:
## Rust vs Bash

> I tackled the rustup codebase and found it too big to use its concepts to make an MVP right now. Was also looking at foundry for inspiration and realized they were using just bash scripts to download and install foundryup, in the end decided to do it in a clap app even though it would probably be much faster if we just script everything, but would probably be harder to introduce features later.

I think using Rust and `clap` is the right choice here. Rust makes it much easier to enable cross-platform support, should we ever want to solve https://github.com/FuelLabs/sway/issues/1526.

The more we lean on bash, the more restricted our options become w.r.t. native Windows support. If we want to support native Windows we would need a batch file equivalent for every bash file we have, no?

I realise we don't publish Windows binaries right now, but it's worth being conscious of these things in order to avoid making https://github.com/FuelLabs/sway/issues/1526 even harder.

## `fuelup` binary

> Included the bin (built for x86_64-apple-darwin) as well for now

Rather than including the binaries as a part of this repo, perhaps we should publish the `fuelup` binary the same way we publish the `forc` binaries under the sway repo using CI? See [here](https://github.com/FuelLabs/sway/blob/master/.github/workflows/ci.yml#L634). Or is this mac binary just temporary for testing purposes?

## Toolchains

> (not sure how this should work - should we be able to set different versions for each of forc, fuel-core, etc?)

Even though it's big, it would be worth looking into how `rustup` does this as it could be tricky to do this nicely in a cross-platform manner.

My initial intuition would be to do something like:

```
~/.fuelup/
    toolchains/
        default # symlink to user-selected default, e.g. -> stable
        nightly # symlink to latest nightly release directory, e.g. -> nightly-2022-05-23
        stable # symlink to latest versioned release directory, e.g. -> v0.13.0
        v0.13.0/
            - forc
            - forc-explore
            - forc-fmt
            - forc-lsp
            - fuel-core
        v0.12.0/
            <binaries>
        nightly-2022-05-23/
            <binaries>
        nightly-2022-05-22/
            <binaries>
```

This way there is a clear, unique location to which we clone each toolchain when we need them, and switching between toolchains is simply a matter of changing the symbolic link.

If it turns out that symbolic links aren't the right choice (e.g. do they even work on Windows? I've never tried), then rather than symbolic links we could a tiny `.rustup/toolchain/toolchain.toml` file or something that points to the current stable, nightly, default and so on.

All that said, there are likely edge cases I'm not thinking of that might influence design, so would recommend trying to drill into rustup at least a bit to see if we're on the right track.

## MVP

> rough minimum viable fuelup that downloads latest core binaries

Fwiw I'd even be happy to just land a barebones/empty CLI app that hardly does anything just so that we can first get all the CI stuff going (just in case there are things with the init script or toolchain support that you're still unsure about or would like reviewed separately).

binggh:
# Rust vs bash

Agreed on your points :) I thought about it for a long time and was considering pushing a shell script to serve as a starter `fuelup` but decided on just starting it off as a rustup and shipping it as a binary right away. Then we can use [`fuelup-init.sh`](https://github.com/FuelLabs/fuelup/pull/8/commits/77af558301ee3f92c5bdaecd6c304cecaf461dd2) for users to download the built version of `rustup` compatible with their system through the script (this is like how rustup-init works as well)

# Fuelup bin

>  Or is this mac binary just temporary for testing purposes?

Yep, I checked it in for now so we can test it - once we have our first proper release with mac/linux compatibility we can remove the binary

# Toolchains

Yeah, I believe what you described was accurate, think I saw a lot of symlink related code within rustup but I will look deeper into it and map it out for myself in a bit.

I think I was wondering more along the lines of how we might want to manage `fuel-core` together with `forc`, seeing as they aren't packaged together and released in 1 version like with packages like `rust-clippy` with `rust`? By the way, I included `fuel-core` in the installation because both are needed to start working with Sway immediately.

I guess rustup allows you to build your own custom toolchain with `rustup toolchain link <toolchain> <path>` that we can follow as well, then if the dev wants to package different versions of binaries together they would do it themselves?

binggh:
Also, should we keep the binaries within `/.cargo/bin` still? Or do we want to have our own bin path

mitchmindtree:
> I think I was wondering more along the lines of how we might want to manage fuel-core together with forc, seeing as they aren't packaged together and released in 1 version like with packages like rust-clippy with rust?

Yeah this is less clear to me too. What happens in the rust + rust-clippy example you gave? How does rustup decide which version to fetch of each?

> I guess rustup allows you to build your own custom toolchain with rustup toolchain link <toolchain> <path> that we can follow as well, then if the dev wants to package different versions of binaries together they would do it themselves?

Yeah it might even be worth looking into solving this first, as it sounds like a more explicit case of handling linking for toolchains more generally. I'd imagine handling `default`, `stable`, `nightly` then might just fall out of this?

> Also, should we keep the binaries within /.cargo/bin still? Or do we want to have our own bin path

Good question! We definitely don't want to edit or touch the user's `.cargo/bin`, and we also don't want to remove anything from the user's `PATH`.

We almost certainly want our own bin path. To ensure that users don't accidentally use a version of one of our binaries from `~/.cargo/bin`, we'll want to have our bin location appear in `PATH` before `~/.cargo/bin` does. Whether we want to edit the `PATH` ourselves or instruct the user to do so manually, I'll defer to following whatever `rustup` does.

binggh:
Testing on this branch, latest commit:

```sh
# curl the script and pass 'install' subcommand
curl -L https://raw.githubusercontent.com/FuelLabs/fuelup/aa08ba4e9c4a13879b802b593c1018917da01da8/fuelup-init.sh | sh -s -- install
```

This installs the latest released versions of `forc`, `forc-fmt`, `forc-lsp`, `forc-explore`, and `fuel-core` into `~/.fuelup`. I'm not quite sure this should be the final directory but perhaps we can store it within `~/.fuelup/bin`, because there will be a need to store metadata related to `fuelup` to allow switching between versions 

Results:

```
Downloading the Forc toolchain
Fetching binary from https://github.com/FuelLabs/sway/releases/download/v0.13.0/forc-binaries-darwin_amd64.tar.gz
Fetching binary from https://github.com/FuelLabs/fuel-core/releases/download/v0.7.0/fuel-core-0.7.0-x86_64-apple-darwin.tar.gz
Unpacking and moving forc to /Users/bh/.fuelup...
Unpacking and moving forc-explore to /Users/bh/.fuelup...
Unpacking and moving forc-lsp to /Users/bh/.fuelup...
Unpacking and moving forc-fmt to /Users/bh/.fuelup...
Unpacking and moving fuel-core to /Users/bh/.fuelup...
```

adlerjohn:
IIRC `rustup` asks the user to modify their `PATH`. Foundry does not and just silently changed my `~/.bashrc`instead of asking and I didn't like that since I put these in `~/.profile` instead of `~/.bashrc`.

binggh:
> IIRC `rustup` asks the user to modify their `PATH`. Foundry does not and just silently changed my `~/.bashrc`instead of asking and I didn't like that since I put these in `~/.profile` instead of `~/.bashrc`.

I think rustup modifies PATH as well no? https://github.com/rust-lang/rustup/blob/master/src/cli/self_update/unix.rs#L82

I think there are installation options you can set for it to not modify your PATH, but the default installation cmd line on rustup.rs has these options:

```
Current installation options:


   default host triple: x86_64-apple-darwin
     default toolchain: stable (default)
               profile: default
  modify PATH variable: yes
```

binggh:
Wait I think I missed the point of your comment, seems like your point is we should probably just have a line telling the user to add to their PATH instead of modifying it for them, in that case I agree

binggh:
Any chance we can merge this to master soon so I can test releases? :) cc: @mitchmindtree 

On that note: is it alright to test releases in this repo or should we use a separate repo for that?

Voxelot:
You can test in another private repo by taking this ci script and adding a beginning step of 

```
      - name: Checkout sources
        uses: actions/checkout@v2
        with:
          repository: 'FuelLabs/fuelup'
          ref: 'binggh/mvp'
```

And then make as many releases as you want without polluting this crate.

src/download.rs
```diff
@@ -0,0 +1,159 @@
...
+use std::fs;
+use std::fs::{File, OpenOptions};
+use std::io::Write;
+use std::time::Duration;
+use tar::Archive;
+
+use crate::constants::FUELUP_PATH;
+
+#[derive(Debug, Serialize, Deserialize)]
+struct LatestReleaseAPIResponse {
```
mitchmindtree, line 17:
```suggestion
struct LatestReleaseApiResponse {
```

src/download.rs
```diff
@@ -0,0 +1,159 @@
...
+    url: String,
+    tag_name: String,
+    name: String,
+}
+
+pub fn forc_bin_tarball_name() -> Result<String> {
+    let os = match std::env::consts::OS {
+        "macos" => Ok("darwin"),
+        "linux" => Ok("linux"),
+        unsupported_os => Err(anyhow!("Unsupported os: {}", unsupported_os)),
```
mitchmindtree, line 27:
```suggestion
        unsupported_os => bail!("Unsupported os: {}", unsupported_os),
```

src/download.rs
```diff
@@ -0,0 +1,159 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+struct LatestReleaseAPIResponse {
+    url: String,
+    tag_name: String,
+    name: String,
+}
+
+pub fn forc_bin_tarball_name() -> Result<String> {
+    let os = match std::env::consts::OS {
+        "macos" => Ok("darwin"),
```
mitchmindtree, line 25:
```suggestion
        "macos" => "darwin",
```

src/download.rs
```diff
@@ -0,0 +1,159 @@
...
+        "linux" => Ok("linux"),
```
mitchmindtree, line 26:
```suggestion
        "linux" => "linux",
```

src/download.rs
```diff
@@ -0,0 +1,159 @@
...
+        unsupported_os => Err(anyhow!("Unsupported os: {}", unsupported_os)),
+    };
+    let architecture = match std::env::consts::ARCH {
+        "aarch64" => Ok("arm64"),
+        "x86_64" => Ok("amd64"),
+        unsupported_arch => Err(anyhow!("Unsupported architecture: {}", unsupported_arch)),
```
mitchmindtree, line 32:
```suggestion
        unsupported_arch => bail!("Unsupported architecture: {}", unsupported_arch),
```

src/download.rs
```diff
@@ -0,0 +1,159 @@
...
+}
+
+pub fn forc_bin_tarball_name() -> Result<String> {
+    let os = match std::env::consts::OS {
+        "macos" => Ok("darwin"),
+        "linux" => Ok("linux"),
+        unsupported_os => Err(anyhow!("Unsupported os: {}", unsupported_os)),
+    };
+    let architecture = match std::env::consts::ARCH {
+        "aarch64" => Ok("arm64"),
```
mitchmindtree, line 30:
```suggestion
        "aarch64" => "arm64",
```

src/download.rs
```diff
@@ -0,0 +1,159 @@
...
+        "x86_64" => Ok("amd64"),
```
mitchmindtree, line 31:
```suggestion
        "x86_64" => "amd64",
```

src/download.rs
```diff
@@ -0,0 +1,159 @@
...
+    };
+
+    Ok(format!("forc-binaries-{}_{}.tar.gz", os?, architecture?))
+}
+
+pub fn fuel_core_bin_tarball_name(version: &str) -> Result<String> {
+    let architecture = match std::env::consts::ARCH {
+        "aarch64" => Ok("aarch64"),
+        "x86_64" => Ok("x86_64"),
+        unsupported_arch => Err(anyhow!("Unsupported architecture: {}", unsupported_arch)),
```
mitchmindtree, line 42: Same as above ^

src/download.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    let vendor = match std::env::consts::OS {
+        "macos" => "apple",
+        _ => "unknown",
+    };
+
+    let os = match std::env::consts::OS {
+        "macos" => Ok("darwin"),
+        "linux" => Ok("linux-gnu"),
+        unsupported_os => Err(anyhow!("Unsupported os: {}", unsupported_os)),
```
mitchmindtree, line 53: Same as above ^

src/download.rs
```diff
@@ -0,0 +1,159 @@
...
+                    // bytes to signal an error condition and return is_write_error.
+                    //
+                    // Reference:
+                    // https://docs.rs/curl/latest/curl/easy/struct.Easy.html#method.write_function
+                    Ok(0)
+                } else {
+                    Ok(new_data.len())
+                }
+            })
+            .unwrap();
```
mitchmindtree, line 133: Should we be returning the error rather than unwrapping?

src/download.rs
```diff
@@ -0,0 +1,159 @@
...
+        transfer.perform().unwrap();
```
mitchmindtree, line 134: Same here ^

src/download.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    let mut data = Vec::new();
+    {
+        let mut transfer = handle.transfer();
+        transfer
+            .write_function(|new_data| {
+                data.extend_from_slice(new_data);
+                Ok(new_data.len())
+            })
+            .unwrap();
```
mitchmindtree, line 82: Error handling?

src/download.rs
```diff
@@ -0,0 +1,159 @@
...
+        transfer.perform().unwrap();
```
mitchmindtree, line 83: Same here ^

mitchmindtree (changes requested):
Just a bunch of small tweaks, but otherwise happy to land this.

Also, rather than using `println!` maybe we should use `tracing` from the start?

binggh:
> Just a bunch of small tweaks, but otherwise happy to land this.
> 
> Also, rather than using `println!` maybe we should use `tracing` from the start?

the `println`s are meant to display info to the end user, is `tracing` still appropriate here?

mitchmindtree:
Yes I think so - you can see some discussion around this for `forc` (which also displays info to the end user) [here](https://github.com/FuelLabs/sway/issues/884#issuecomment-1124524745).

src/commands/install.rs
```diff
@@ -27,6 +28,9 @@ pub fn install() -> Result<()> {
         GITHUB_API_REPOS_BASE_URL, FUEL_CORE_REPO, RELEASES_LATEST
     ))?;
 
+    info!("Fetching forc {}", &forc_release_latest_tag);
+    info!("Fetching fuel-core {}", &fuel_core_release_latest_tag);
```
mitchmindtree, line 32: Should we move this to after `forc` is fetched?

mitchmindtree: Ohhh my apologies, I thought these were meant to indicate when the fetching of each package begins, so I thought each of these should happen before their corresponding download begins.

mitchmindtree (approved):

PR #9, opened 2022-05-23
preliminary ci.yml

binggh:


Voxelot:
Maybe target this PR into the other one, so that we know the initial PR to master will pass CI.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,135 @@
...
+        if: ${{ matrix.job.target == 'aarch64-unknown-linux-gnu' }}
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install -y gcc-aarch64-linux-gnu
+          echo "CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc" >> $GITHUB_ENV
+      - name: Install Fuelup
+        uses: actions-rs/cargo@v1
+        with:
+          command: install
+          args: --profile=release --path .
```
Voxelot, line 115: Just a heads up that this may not work as expected if you ever use rust-bindgen. Cross is the more standard approach, take a look at [ripgrep](https://github.com/BurntSushi/ripgrep/blob/master/.github/workflows/release.yml) for example

binggh: Thanks for the heads up and the link - updated the CI to use `cross` instead

.github/workflows/ci.yml
```diff
@@ -0,0 +1,156 @@
...
+        run: |
+          echo "SDKROOT=$(xcrun -sdk macosx --show-sdk-path)" >> $GITHUB_ENV
+          echo "MACOSX_DEPLOYMENT_TARGET=$(xcrun -sdk macosx --show-sdk-platform-version)" >> $GITHUB_ENV
+
+      - name: Linux ARM setup
+        if: ${{ matrix.job.target == 'aarch64-unknown-linux-gnu' }}
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install -y gcc-aarch64-linux-gnu
+          echo "CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc" >> $GITHUB_ENV
```
Voxelot, line 118: cross handles all the sysroot config and more

Voxelot, line 118: cross handles all the sysroot config and more

```suggestion
```

binggh:
Realise there's also an issue with the final step, shouldn't be able to do `which fuelup` anymore since we're not installing it like in [sway CI](https://github.com/FuelLabs/sway/blob/master/.github/workflows/ci.yml)

Voxelot:
This is how I did it without using `which`: https://github.com/FuelLabs/fuel-core/blob/master/.github/workflows/ci.yml#L356

binggh:
Hmm, now testing this on [a branch within the fuel-dummy-test repo](https://github.com/FuelLabs/fuel-dummy-test/tree/binggh/fuelup-binary-test) and the cross images are running into [build errors](https://github.com/FuelLabs/fuel-dummy-test/runs/6575781790?check_suite_focus=true) which seem to stem from OpenSSL. Seems to be a long-standing issue and point of discussion over at cross: https://github.com/cross-rs/cross/issues/229

[Workaround](https://github.com/cross-rs/cross/issues/229#issuecomment-748500115) seems to be to use a custom image and install the necessary packages(?) Trying this now.

Voxelot:
Ah interesting, it may be worth investigating a way to eliminate the dependency on OpenSSL given most libraries support webpki/rustls these days. I did have to do custom docker images for fuel-core but it was painful.

binggh:
Ah, okay 👍  Seems possible to build with `rustls` instead so I think I'll work towards that, thanks @Voxelot!

binggh:
> Ah, okay 👍 Seems possible to build with `rustls` instead so I think I'll work towards that, thanks @Voxelot!

Turns out `curl-rust` allows disabling openssl features and enabling rustls instead ([link](https://github.com/alexcrichton/curl-rust/blob/main/Cargo.toml#L47)):

```
curl = { version = "0.4" default-features = false, features = ["rustls"] }
```

Builds seem to be working fine in the test repo.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,150 @@
...
+        if: matrix.job.target == 'aarch64-unknown-linux-gnu'
+        run: |
+          docker run --rm -v \
+          "$PWD/target:/target:Z" \
+          aarch64-linux-gnu:latest \
+          aarch64-linux-gnu-strip \
+          /target/aarch64-unknown-linux-gnu/release/fuelup
+
+      - name: Strip release binary mac
+        if: matrix.job.os == 'mac-os'
```
Voxelot, line 114: ah found a bug in the fuel-core script as well

```suggestion
        if: matrix.job.os == 'macos-latest'
```

mitchmindtree (approved):
Lgtm!

mitchmindtree (changes requested):
Oh, pending @Voxelot's fix - good spot

.github/workflows/ci.yml
```diff
@@ -0,0 +1,150 @@
...
+      - name: Strip release binary x86_64-linux-gnu
+        if: matrix.job.target == 'x86_64-unknown-linux-gnu'
+        run: strip "target/${{ matrix.job.target }}/release/fuelup"
+
+      - name: Strip release binary aarch64-linux-gnu
+        if: matrix.job.target == 'aarch64-unknown-linux-gnu'
+        run: |
+          docker run --rm -v \
+          "$PWD/target:/target:Z" \
+          aarch64-linux-gnu:latest \
```
Voxelot, line 109:
```suggestion
          ghcr.io/cross-rs/aarch64-unknown-linux-gnu:main \
```

Voxelot:
should be g2g, but let's see a test release fully work before merging

binggh:
Is there supposed to be some way to verify the developer(s) of the releases? Couldn't run the x86_64 mac binary myself at first, then tried a bit later and it worked somehow

Voxelot:
We'll need to get an apple developer certificate enabled in CI to resolve that. For now we'll need to find some kind of workaround to help users requests the permissions they need to run the binary on mac.

Voxelot (approved):

adlerjohn:
If it's not needed for this PR, can you file an issue and assign it to me to follow-up on adding the cert?

PR #11, opened 2022-05-25
Remove fuelup directory and bin which was for testing purposes

binggh:


Voxelot (approved):
do we need a git ignore for this?

binggh:
> do we need a git ignore for this?

Nope - i moved it out of `target` because i wanted to be able to `curl` from it to test

adlerjohn (approved):

PR #14, opened 2022-05-26
rework fuelup-init.sh

binggh:
There's a bunch of changes here, but essentially the main changes are within e42b8c18f51b0c22ddc1d62fc593ab648189bf3c, 4ab5912740b69c4b50a40039addf09c763ba0e6f

### How this works
- script will detect arch of the user running this script, download the appropriate release of fuelup into a `/tmp` folder, extract and move it into `.fuelup/bin`. Previously it didn't, because initial script was basically `rustup-init` but they do an intermediate step in between, which we don't at the moment.

Once this is merged and a new release is out we should be able do something like `curl <url> |sh -s install` and it will auto run `fuelup install` in a single step


fuelup-init.sh
```diff
@@ -1,42 +1,40 @@
...
+    need_cmd rm
+    need_cmd rmdir
+
+    get_architecture || return 1
+    local _arch="$RETVAL"
+    assert_nz "$_arch" "arch"
+
+    mkdir -p "$FUELUP_DIR/bin"
+
+    local _fuelup_version="0.0.2"
```
adlerjohn, line 20: Instead, can you get the latest tagged release? ref: https://github.com/celestiaorg/docs/blob/d73cc3cf571040e9b589b5f194e8fb73f9e7b452/docs/developers/celestia-app.md?plain=1#L18-L20

src/commands/install.rs
```diff
@@ -74,5 +76,8 @@ pub fn install() -> Result<()> {
         }
     }
 
+    info!("The Forc toolchain is installed now. Great!\n");
```
adlerjohn, line 79: Maybe say what specific version was installed?

adlerjohn (approved):
utACK, but will defer to @mitchmindtree for final review.

src/constants.rs
```diff
@@ -1,4 +1,5 @@
 pub const FUELUP_PATH: &str = ".fuelup";
```
mitchmindtree, line 1:
```suggestion
pub const FUELUP_DIR: &str = ".fuelup";
```

Maybe worth clarifying that this is the directory name to avoid confusion with fuelup exe path?

src/constants.rs
```diff
@@ -1,4 +1,5 @@
...
+pub const FUELUP_BIN_PATH: &str = ".fuelup/bin";
```
mitchmindtree, line 2: I realise we already have a bash script which is unix-specific, but let's not also hard-code unix-style paths if we can avoid it. Instead let's use a `fuelup_bin_dir` function that uses `Path::join` or something like that.

src/commands/install.rs
```diff
@@ -35,6 +37,12 @@ pub fn install() -> Result<()> {
         Err(_) => bail!("Failed to fetch latest fuel-core release tag from GitHub API"),
     };
 
+    let fuelup_path = home_dir().unwrap().join(Path::new(FUELUP_BIN_PATH));
```
mitchmindtree, line 40: Let's re-use the `fuelup_bin_dir` function here, and maybe change the `fuelup_path` binding to reflect it.

src/commands/install.rs
```diff
@@ -52,15 +60,14 @@ pub fn install() -> Result<()> {
...
-    let fuelup_path = home_dir().unwrap().join(Path::new(FUELUP_PATH));
     for entry in std::fs::read_dir(&fuelup_path)? {
         let sub_path = entry?.path();
 
         if sub_path.is_dir() {
             for bin in std::fs::read_dir(&sub_path)? {
                 let bin_file = bin?;
                 info!(
-                    "Unpacking and moving {} to {}/.fuelup",
+                    "Unpacking and moving {} to {}/.fuelup/bin",
```
mitchmindtree, line 70: We should use `fuelup_path.display()` rather than manually writing the path.

src/commands/install.rs
```diff
@@ -74,5 +81,12 @@ pub fn install() -> Result<()> {
         }
     }
 
+    info!(
+        "\n\nInstalled: forc {}, fuel-core {}",
+        forc_release_latest_tag, fuel_core_release_latest_tag
+    );
+    info!("\nThe Forc toolchain is installed now. Great!");
+    info!("\nYou might need to add $HOME/.fuelup/bin to your path.");
```
mitchmindtree, line 89: Same with the path here ^

mitchmindtree (changes requested):
Lots of indentation changes to sift through, but otherwise lgtm! Just a couple small changes.

src/commands/install.rs
```diff
@@ -37,10 +35,8 @@ pub fn install() -> Result<()> {
         Err(_) => bail!("Failed to fetch latest fuel-core release tag from GitHub API"),
     };
 
-    let fuelup_path = home_dir().unwrap().join(Path::new(FUELUP_BIN_PATH));
-
-    if !fuelup_path.is_dir() {
-        fs::create_dir_all(&fuelup_path)?;
+    if !fuelup_bin_dir().is_dir() {
```
mitchmindtree, line 36:
```suggestion
    let fuelup_bin_dir = fuelup_bin_dir();
    if !fuelup_bin_dir.is_dir() {
```
Constructing `PathBuf` performs dynamic allocation (much like Vec and String). It doesn't matter much in this case but as general good practise it would be better to construct it once and re-use it when it's trivial to do so.

mitchmindtree (approved):
Looks good, one last tip but it's not a blocker.

mitchmindtree (approved):

PR #22, opened 2022-05-27
Add fuelup installation message at the end

binggh:
Add a simple message to let users know that `fuelup` has been installed and that they can use `fuelup install` to fetch the latest binaries

### DIFF
--- a/fuelup-init.sh
+++ b/fuelup-init.sh
@@ -60,7 +60,7 @@ main() {
     ignore rmdir "$_dir/fuelup-${_fuelup_version}-${_arch}"
     ignore rmdir "$_dir"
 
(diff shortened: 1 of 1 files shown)


fuelup-init.sh
```diff
@@ -60,6 +60,8 @@ main() {
     ignore rmdir "$_dir/fuelup-${_fuelup_version}-${_arch}"
     ignore rmdir "$_dir"
 
+    echo "\nfuelup ${_fuelup_version} has been installed in $FUELUP_DIR/bin. To fetch the latest forc and fuel-core binaries, you can do 'fuelup install' in future."
```
adlerjohn, line 63:
```suggestion
    echo "\nfuelup ${_fuelup_version} has been installed in $FUELUP_DIR/bin. To fetch the latest forc and fuel-core binaries, run 'fuelup install'."
```

adlerjohn (approved):

PR #30, opened 2022-05-28
deprecate curl in favour of ureq

### LINKED ISSUE #29: Use `ureq` instead of `curl` with `rustls`
`curl` with `rustls` is unstable, see: https://github.com/alexcrichton/curl-rust#building

> rustls Enable SSL/TLS support via [Rustls](https://github.com/ctz/rustls), a well-received alternative TLS backend written in Rust. Rustls is always statically linked. Disabled by default.

Consider using `ureq` instead as a native-Rust alternative. Ref: https://github.com/FuelLabs/sway/pull/651

binggh:
closes #29

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,6 @@ path = "src/main.rs"
 [dependencies]
 anyhow = "1"
 clap = { version = "3.1", features = ["cargo", "derive", "env"] }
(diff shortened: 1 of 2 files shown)


adlerjohn (approved):
utACK

PR #33, opened 2022-05-28
commit Cargo.lock

### LINKED ISSUE #32: Commit lockfile
`fuelup` is a binary, so I think the lockfile should be committed. cc @mitchmindtree

binggh:
Closes #32 

Commit Cargo.lock, and remove Cargo.lock from .gitignore


mitchmindtree:
Can you also add `--locked` to the relevant CI pass if possible? I think it might only be clippy that needs it. This will ensure the lock file stays up to date on master in the case a PR would edit `[dependencies]`.

mitchmindtree (approved):

adlerjohn (approved):

PR #38, opened 2022-05-28
Fix script location.

adlerjohn:


binggh:
technically shouldn't we keep `verify_tags.sh` in `/scripts` instead and change the CI instead?

adlerjohn:
Maybe, but I just kept it consistent with how it is in every other repo.

binggh (approved):
Right, I was looking at Sway repo as well as it had a separate `/scripts` repo 😂 so I guess this works

Issue #31, opened 2022-05-28
Use `shellcheck` in CI to ensure `fuelup-init.sh` reliability


binggh:
https://github.com/koalaman/shellcheck
https://github.com/ludeeus/action-shellcheck

Could be useful to ensure reliability of `fuelup-init.sh`

Edit: this should include a formatter as well.

adlerjohn:
This is exactly what I used to clean up in #28!

binggh:
TIL of this useful tool

mitchmindtree:
TIL too, sounds like a great idea!

PR #48, opened 2022-05-30
Update err message to handle failed download

### LINKED ISSUE #44: `fuelup-init.sh` should fail gracefully if binary can't be found
In between publishing a release and binaries being built and uploaded by CI, a user that runs `fuelup-init.sh` will be greeted with a message like

```console
info: downloading installer
curl: (22) The requested URL returned error: 404
fuelup: installer for platform 'aarch64-apple-darwin' not found, this may be unsupported
```

`fuelup-init.sh` should fail gracefully and warn the user that they might be in this situation, and try again in a few minutes. Maybe provide a link to the actual release page on GitHub too so they can manually confirm there are release binaries, and the time of release, etc.

binggh:
Closes #44 

- Add `_fuelup_version` to the info printed, so that users can check versioning at the release page
- Clearer err message so that users know what went wrong and where to check instead of just 'unsupported arch' -> actually, in the case of 'unsupported arch', that check should've failed at `get_architecture` already
- also removed the `$` matching in `grep` - i'm not sure why matching `$` didn't work on my machine, but matching `404` should be sufficient either way since there would be nowhere else in the message that should have a `404`

### DIFF
--- a/fuelup-init.sh
+++ b/fuelup-init.sh
@@ -227,7 +227,7 @@ downloader() {
         if [ -n "$_err" ]; then
             echo "$_err" >&2
             if echo "$_err" | grep -q 404; then
-		err "fuelup was not found - either the release is not ready yet or the tag is invalid. You can check if the release is available here: https://github.com/FuelLabs/fuelup/releases"
+		err "fuelup ${_fuelup_version} was not found - either the release is not ready yet or the tag is invalid. You can check if the release is available here: https://github.com/FuelLabs/fuelup/releases/${_fuelup_version}"
             fi
         fi
 
--- a/fuelup-init.sh
+++ b/fuelup-init.sh
@@ -79,9 +79,9 @@ main() {
     done
 
     if $_ansi_escapes_are_valid; then
-        printf "\33[1minfo:\33[0m downloading installer\n" 1>&2
+        printf "\33[1minfo:\33[0m downloading fuelup %s\n" $_fuelup_version 1>&2
     else
-        printf '%s\n' 'info: downloading installer' 1>&2
+        printf 'info: downloading fuelup %s\n' $_fuelup_version 1>&2
     fi
 
     ensure downloader "$_fuelup_url" "$_file" "$_arch"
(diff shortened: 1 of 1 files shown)


fuelup-init.sh
```diff
@@ -226,8 +226,8 @@ downloader() {
         fi
         if [ -n "$_err" ]; then
             echo "$_err" >&2
-            if echo "$_err" | grep -q 404$; then
-                err "installer for platform '$3' not found, this may be unsupported"
+            if echo "$_err" | grep -q 404; then
+		err "fuelup was not found - either the release is not ready yet or the tag is invalid. You can check if the release is available here: https://github.com/FuelLabs/fuelup/releases"
```
mitchmindtree, line 230: Indentation?

mitchmindtree (dismissed):
Lgtm, just an optional indentation fix

fuelup-init.sh
adlerjohn, line 230: Link to the actual release that's being tried

binggh: are spaces preferred here? I ran it all under a shell formatter one of the last PRs. Can we disregard this indentation issue this PR and I will open another PR running `shfmt`? Would probably be worth including in #31 as well

binggh: Hmm, I linked to the releases page as a conscious decision since linking to a version that doesn't exist results simply in a Github 404 page, which isn't very informative (imo). Linking to the releases page allows the user to look at the latest release from the top anyway, and to manually check if the version the script was trying to download exists.

With that said, will update it to link to the actual release though - if the above logic makes sense I can change it back in future

adlerjohn: Unless something goes very wrong, the only time this error should be hit under normal operation is when there's a release but not binaries uploaded yet. So not a 404. If it's a 404 then that means something went really wrong, for which linking to the exact page is probably more informative.

adlerjohn (approved):
utACK, but will defer to @mitchmindtree for final review.

mitchmindtree (approved):

PR #50, opened 2022-05-30
replace jq with cut

### LINKED ISSUE #52: Replace `jq` with `cut` as an intermediate solution to remove dependence on `jq`
#50 has replaced `jq` with `cut`, but it would still be ideal to entirely remove dependence on `jq` as mentioned in #43 

Ideally a non-brittle solution would be preferred.

binggh:
Closes #52 

We use a combination of `grep` and `cut` instead of `jq`.

`"tag_name": "v0.1.2"`

The above is the output of the grep, then:

`cut -d "\"" -f4 | cut -c 2-`

this splits the above string with the `"` delimiter, getting the `v0.1.2` portion and cutting off the `v`

This may not also be the ideal way, but it's one solution as well, unless the solution described in the issue is preferred.

### DIFF
--- a/fuelup-init.sh
+++ b/fuelup-init.sh
@@ -24,7 +24,7 @@ main() {
     mkdir -p "$FUELUP_DIR/bin"
 
     local _fuelup_version
-    _fuelup_version="$(curl -s https://api.github.com/repos/FuelLabs/fuelup/releases/latest | grep -r "tag_name" | cut -d "\"" -f4 | cut -c 2-)"
+    _fuelup_version="$(curl -s https://api.github.com/repos/FuelLabs/fuelup/releases/latest | grep "tag_name" | cut -d "\"" -f4 | cut -c 2-)"
     local _fuelup_url="https://github.com/FuelLabs/fuelup/releases/download/v${_fuelup_version}/fuelup-${_fuelup_version}-${_arch}.tar.gz"
 
     local _dir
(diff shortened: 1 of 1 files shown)


fuelup-init.sh
```diff
@@ -25,8 +24,7 @@ main() {
     mkdir -p "$FUELUP_DIR/bin"
 
     local _fuelup_version
-    _fuelup_version="$(curl -s https://api.github.com/repos/FuelLabs/fuelup/releases/latest | jq -r ".tag_name")"
-    _fuelup_version="$(echo "${_fuelup_version}" | cut -c 2-)"
+    _fuelup_version="$(curl -s https://api.github.com/repos/FuelLabs/fuelup/releases/latest | grep -r "tag_name" | cut -d "\"" -f4 | cut -c 2-)"
```
mitchmindtree, line 27:
```suggestion
    _fuelup_version="$(curl -s https://api.github.com/repos/FuelLabs/fuelup/releases/latest | grep "tag_name" | cut -d "\"" -f4 | cut -c 2-)"
```
Just tested this locally, but had to remove the recursive flag to get this to work. Otherwise looks good!

mitchmindtree (changes requested):
TIL about cut

adlerjohn (commented):
Can you file an issue to temporary-workaround replacing `jq`, and close that instead? We should keep the non-brittle solution as an open issue.

adlerjohn (approved):
utACK

mitchmindtree (approved):

PR #51, opened 2022-05-30
`fuelup self update`

binggh:
Allows users to run `fuelup self update` to download the latest `fuelup` into `.fuelup/bin`

### DIFF
--- a/src/commands/fuelup.rs
+++ b/src/commands/fuelup.rs
@@ -3,14 +3,13 @@ use clap::Parser;
 use tracing::info;
 
 use crate::{
-    commands::install::unpack_extracted_bins,
     constants::{FUELUP_RELEASE_DOWNLOAD_URL, GITHUB_API_REPOS_BASE_URL, RELEASES_LATEST},
-    download::{download_file_and_unpack, fuelup_bin_dir, get_latest_tag},
+    download::{download_file_and_unpack, fuelup_bin_dir, get_latest_tag, unpack_extracted_bins},
 };
 
 #[derive(Debug, Parser)]
 pub enum FuelupSubcommand {
-    /// Updates the fuelup tool.
+    /// Updates fuelup
     Update,
 }
 
@@ -35,8 +34,6 @@ fn fuelup_bin_tarball_name(version: &str) -> Result<String> {
         unsupported_os => bail!("Unsupported os: {}", unsupported_os),
     };
 
-    println!("{:?}", version);
-
     Ok(format!(
         "fuelup-{}-{}-{}-{}.tar.gz",
         // strip the 'v' from the version string to match the file name of the releases
@@ -48,12 +45,6 @@ fn fuelup_bin_tarball_name(version: &str) -> Result<String> {
 }
 
 pub fn self_update() -> Result<()> {
-    // get latest fuelup tag
-    //
-    // download to a tmp directory
-    //
-    // replace with current fuelup
-    //
     let proc = std::process::Command::new("fuelup")
         .arg("--version")
         .output()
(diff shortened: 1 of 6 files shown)


adlerjohn (commented):
Looks fine to me, will defer to @mitchmindtree

binggh:
I would like to at some point implement what [cargo](https://github.com/rust-lang/cargo/blob/master/src/cargo/ops/cargo_install.rs#L276) does handling installing/reinstalling binaries now that we have some duplication in code from `download.rs`. The current mechanism might be a bit flimsy in terms of error handling especially when unexpected errors occur, my main concern is leaving unwanted files/directories behind in the user's filesystem (which is intrusive and isn't nice). It might be too big for this PR and will come in the form of a refactor PR.

src/commands/fuelup.rs
```diff
@@ -0,0 +1,81 @@
...
+use clap::Parser;
+use tracing::info;
+
+use crate::{
+    constants::{FUELUP_RELEASE_DOWNLOAD_URL, GITHUB_API_REPOS_BASE_URL, RELEASES_LATEST},
+    download::{download_file_and_unpack, fuelup_bin_dir, get_latest_tag, unpack_extracted_bins},
+};
+
+#[derive(Debug, Parser)]
+pub enum FuelupSubcommand {
```
mitchmindtree, line 11:
```suggestion
pub enum FuelupCommand {
```
^ to be consistent with `InstallCommand` maybe?

src/commands/fuelup.rs
```diff
@@ -0,0 +1,81 @@
...
+    Update,
+}
+
+#[derive(Debug, Parser)]
+struct UpdateCommand {}
+
+fn fuelup_bin_tarball_name(version: &str) -> Result<String> {
+    let architecture = match std::env::consts::ARCH {
+        "aarch64" => "aarch64",
+        "x86_64" => "x86_64",
```
mitchmindtree, line 22: We can handle the supported cases in one branch:

```rust
"aarch64" | "x86_64" => std::env::consts::ARCH,
```

src/commands/fuelup.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+pub fn self_update() -> Result<()> {
+    let proc = std::process::Command::new("fuelup")
+        .arg("--version")
+        .output()
+        .expect("Could not run fuelup.");
+
+    let stdout = String::from_utf8_lossy(&proc.stdout);
+    let current_version = stdout.split_whitespace().nth(1).unwrap();
```
mitchmindtree, line 54:
```suggestion
    let current_version = stdout.split_whitespace().nth(1).expect("failed to parse current version from `fuelup --version` output");
```

src/commands/fuelup.rs
```diff
@@ -0,0 +1,81 @@
...
+        vendor,
+        os
+    ))
+}
+
+pub fn self_update() -> Result<()> {
+    let proc = std::process::Command::new("fuelup")
+        .arg("--version")
+        .output()
+        .expect("Could not run fuelup.");
```
mitchmindtree, line 51: Is there any way to access the version of the current `fuelup` clap app instance that's running without re-invoking it using `Command`?

mitchmindtree (commented):
Nice, mostly lgtm!

> my main concern is leaving unwanted files/directories behind in the user's filesystem (which is intrusive and isn't nice), or deleting the wrong files when cleaning up.

I wouldn't worry about this too much just yet. I imagine having multiple toolchains installed (e.g. nightly, stable) will affect what we consider to be "unused" bins. E.g. when using rustup if you use stable, then switch to nightly, and then switch back to stable I believe it will still have the stable toolchain "cached" (and doesn't need to redownload it), even though it wasn't currently in use.

src/commands/fuelup.rs
mitchmindtree: Answering my own question as I remembered we already do this in forc in order to work out the sway git tag there. Check out this macro:

https://docs.rs/clap/latest/clap/macro.crate_version.html

src/commands/fuelup.rs
binggh: Hmmm, I was thinking about the naming of this for a bit. What I wanted to achieve was to separate a `Command` without additional `Command`s but now that I look at it again it does seem wrong that it is called a `Subcommand`... I think perhaps I'll indeed change it back so that it's consistent with `forc`

src/commands/fuelup.rs
binggh: this is super helpful, cut down quite a lot of code from this

mitchmindtree (approved):

PR #56, opened 2022-05-31
bootstrap ci tests

### LINKED ISSUE #2: Add CI for building, testing, publishing
Once we've got a bare-bones cargo project, we should prioritise setting up CI.

We'll likely want to publish the `fuelup` binary for each platform similar to how we publish `forc` for each platform too.

binggh:
Should close #2 for good, because testing is the only missing step.

- bootstrapping with a simple test just so we can get a test CI up and running.
- don't think other areas of the code are valid for testing now, perhaps we might want a more comprehensive test suite (i like cargo's one more than rustup's) once we get more commands up and running

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,22 @@ jobs:
           command: fmt
           args: --all -- --check
 
+  cargo-test:
+    needs: cancel-previous-runs
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+      - name: Run tests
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -50,6 +50,22 @@ jobs:
...
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+      - name: Run tests
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --locked
+
```
adlerjohn, line 67:
```suggestion
          args: --locked
```

adlerjohn: Er. Delete newline

adlerjohn (approved):

PR #58, opened 2022-06-01
ci for deploying fuelup version

### LINKED ISSUE #43: Remove dependence on `jq` in `fuelup-init.sh`
`fuelup-init.sh` currently depends on `jq` to parse the JSON response from GitHub for the latest release tag of `fuelup`. It would be nice to not have to depend on `jq`.

https://github.com/FuelLabs/fuelup/blob/69b955cc1545375f6f40ce8c406299e224c9b17c/fuelup-init.sh#L28

A pure-shell implementation would be fragile and so is discouraged. An alternative would be to, on release, publish a single file to the `gh-pages` branch that contains the latest release tag. Then it can be downloaded with `curl` (which is already an unavoidable dependency).

binggh:
Closes #43 

We publish a single file to the gh-pages branch that contains the latest release tag, as suggested in the issue.


.github/workflows/gh-pages-init.yml
```diff
@@ -27,10 +28,25 @@ jobs:
...
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           keep_files: true
           publish_dir: ${{ env.FUELUP_DIR }}
           destination_dir: ./
+
+      - name: Get latest tag
+        run: |
+          echo ${GITHUB_REF#refs/tags/} >> fuelup-version
+          cp fuelup-version ${{ env.FUELUP_VERSION_DIR }}
```
adlerjohn, line 42: You might have to `mkdir -p` before this step.

.github/workflows/gh-pages-init.yml
```diff
@@ -27,10 +28,25 @@ jobs:
...
+        if: startsWith(github.ref, 'refs/tags')
```
adlerjohn, line 43: You don't need this because this job only runs on release publishing.

.github/workflows/gh-pages-init.yml
```diff
@@ -27,10 +28,25 @@ jobs:
...
+
+      - name: Deploy latest fuelup version
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          keep_files: true
+          publish_dir: ${{ env.FUELUP_VERSION_DIR }}
+          destination_dir: ./
+        if: startsWith(github.ref, 'refs/tags')
```
adlerjohn, line 52: Ditto

adlerjohn (commented):
Maybe rename the file to something like `gh-pages-meta.yml`?

.github/workflows/gh-pages-init.yml
binggh: good point

binggh:
tested the action on a private repo, seems to be good to go 👍  the final diff is uninformative because of the renaming but please check the 4 commits before [ccb672d](https://github.com/FuelLabs/fuelup/pull/58/commits/ccb672d73ca21995c2b36110b3c1ac216479f123) for reviewing!

adlerjohn (approved):
Looks good to me, but will defer final review to @mitchmindtree

mitchmindtree (approved):

Issue #63, opened 2022-06-07
Support the toolchain pattern, starting with `latest` toolchain


binggh:
`fuelup` is largely inspired by `rustup` in terms of how it should distribute packages in the Fuel toolchain in future, and this inspiration extends to the [toolchain concept](https://rust-lang.github.io/rustup/concepts/toolchains.html).

Following a discussion [here](https://github.com/FuelLabs/fuelup/issues/46#issuecomment-1143790506), a good first step towards building the foundation of toolchain management would be to support the `latest` toolchain (which is basically what we have now!)

This would mean:
- We have some kind of config file within `~/.fuelup/bin`, this could be in the form of a `settings.toml` similar to rustup, where configs (eg. the default toolchain) are set,
- We download bins to `/toolchains/latest/bin` instead, so that next time we can download other versions to, for example, `/toolchains/stable/bin`, or we can even support custom toolchains i.e. `/toolchains/my-toolchain/bin`

binggh:
In making this work we need to support the [proxy pattern](https://rust-lang.github.io/rustup/concepts/proxies.html) - it means we will run `forc` and `fuel-core` through `fuelup`, pass necessary args through `fuelup` to the respectively tools. This is how `rustup` is aware of which toolchain to use for which tool.

PR #64, opened 2022-06-07
`fuelup check`

### LINKED ISSUE #16: `fuelup check`: command to check for updates
Having a `fuelup check` command that tells us if we need to update our toolchain would be useful. Although this command is technically not needed (because `fuelup install` downloads the latest binaries anyway), this command would be helpful in branching out into building other commands like `fuelup update` (updates all components), `fuelup default [stable|nightly]` (switch between toolchains), since these commands rely on versioning checks or toolchain checks (which we do not have the concept of yet).

~We can store update tags within a dedicated folder in `.fuelup`, in the format `[bin]-[version]` (like how our packages are released at the moment):~

```
/.fuelup
    /bin
        forc
        ...
    /binary-versions
        fuel-core-0.8.0
```

~Then we check if the tag version is the same. Simple solution for now until we have toolchain versions in place - then it's also trivial to switch over to checking for the toolchain version~

Basic implementation is now at #64 - for now we are running whichever component is first available from path atm. In future there is probably a better way to do this, especially when we have multiple toolchains.

binggh:
Closes #16 
(note: this merges into #62)

Now that we have `DownloadCfg` in #62 it's much easier to support `fuelup check`, since all we need to do is to build the cfg and compare it with the current version running on the machine.

## Usage

```console
$ fuelup check
forc - update available: v0.14.5 -> v0.15.1
fuel-core - up to date: v0.8.0
fuelup - up to date: v0.1.2
```

~There is definitely a more elegant way of handling checks - i.e. we shouldn't panic on each command but simply collect the results and print it all at the end. For example, if `fuel-core` was not installed, `fuelup check` would panic, but ideally it should do something like this~:

Have since updated this PR by not panicking if the component cannot be found/run, instead showing a `not found` message. I was wondering if there should be an action message here i.e. if any of `forc`, `forc-fmt`, `forc-explore`, or `forc-lsp` are missing or outdated, should we add a reminder to do `fuelup install`? Then again it might be intentionally not updated on purpose (if say `forc` has a breaking change that breaks a dev's project). Due to this reason I decided not to include this action message.

Also included plugins in the check.

```console
$ fuelup check
forc - up to date: v0.15.1
 - forc-fmt - up to date: v0.15.1
 - forc-explore - update available: v0.12.2 -> v0.15.1
 - forc-lsp not found
fuel-core - up to date: v0.8.0
fuelup - up to date: v0.1.2
```


binggh:
I added in checks for plugins as well:

```console
$ fuelup check
forc - up to date: v0.15.1
 - forc-fmt - up to date: v0.15.1
 - forc-explore - update available: v0.12.2 -> v0.15.1
 - forc-lsp not found
fuel-core - up to date: v0.8.0
fuelup - up to date: v0.1.2
```

Brings up the question of how we should treat the plugins packaged together with `forc` - should we consider the plugins shipped with `forc` to just be `forc` as well? I would assume so and hence in this implementation I indent the checks for the plugins under the check for `forc`.

binggh:
Converting back to draft for now since changes from #62 need to be merged here

binggh:
Updates:

- Updated this command to use `semver::Version` following changes to #62 
- Updated usages of components to use a struct `component` containing these constants instead
- Moved plugin constants into a similar setup as above

adlerjohn (commented):
Rustup uses a specific dark green and yellow for up to date and outdated. It also uses bold for the non-version text.

```console
$ rustup check
stable-x86_64-unknown-linux-gnu - Up to date : 1.61.0 (fe5b13d68 2022-05-18)
nightly-x86_64-unknown-linux-gnu - Up to date : 1.63.0-nightly (5435ed691 2022-06-07)
rustup - Up to date : 1.24.3
```

Above, the text `stable-x86_64-unknown-linux-gnu - Up to date` is in bold and `: 1.61.0 (fe5b13d68 2022-05-18)` is not bold. `Up to date` is green.

Can you use exactly the same format?

binggh:
Converting this to a draft and marking this as blocked until #63 is cleared.

binggh:
Here's what I came up with with the presentation of `fuelup check` - this is different from `rustup check` because we want to show `forc` and `fuel-core` versions separately within the same toolchain (rustc, cargo, etc. all share the same versioning, which doesn't work for us):

Screenshot 2022-06-19 at 11 11 18 PM

It's a bit verbose/long, especially if we have multiple toolchains, since `fuelup check` lists all toolchains under `/toolchains` and checks the versioning for all of them.

This can be shortened however - since the plugin binaries are still packaged together with `forc` we don't have to show the versioning for the plugins. So we can do something like this as well:

Screenshot 2022-06-19 at 11 14 22 PM

Or if we want to show slightly more info:

Screenshot 2022-06-19 at 11 18 17 PM

I think the last one has the best balance between showing enough info and not stretching the output vertically too much. I'd pick either the last or the first one.

P.S.: I realized after taking the screenshots that I was missing a `:` after `Update available`, please pretend it's there 😅

adlerjohn:
Last one looks good. I'm actually leaning towards the second since the information on the bundled plugins takes up a lot of horizontal real estate. The first definitely has too much redundancy.

binggh:
> Last one looks good. I'm actually leaning towards the second since the information on the bundled plugins takes up a lot of horizontal real estate. The first definitely has too much redundancy.

Agreed on the last one taking too much horizontal real estate as well, leaning towards the second one myself. What about not specifying that they're plugins and just listing them instead?

Screenshot 2022-06-20 at 9 41 25 AM

With that said I think I like the 2nd one more because it's short and sweet and it's implicit that the plugins are packaged with forc anyway, so I'm updating it to that

adlerjohn:
Maybe add a more verbose option that explicitly lists bundled plugins?

src/colors.rs
```diff
@@ -0,0 +1,30 @@
...
+use termcolor::{Color, ColorChoice, ColorSpec, StandardStream, WriteColor};
+
+pub fn print_boldln(txt: &str) -> Result<()> {
+    let mut stdout = StandardStream::stdout(ColorChoice::Auto);
+    stdout.set_color(ColorSpec::new().set_bold(true))?;
+
+    writeln!(&mut stdout, "{}", txt)?;
+    stdout.reset()?;
+    Ok(())
+}
```
mitchmindtree, line 12: For a bit more flexibility w.r.t. formatting etc, you could do something like:

```rust
pub fn bold<F>(write: F) -> Result<()>
where
    F: FnOnce(&mut StandardStream),
{
    let mut stdout = StandardStream::stdout(ColorChoice::Auto);
    stdout.set_color(ColorSpec::new().set_bold(true))?;
    write(&mut stdout)?;
    stdout.reset()?;
    Ok(())
}
```
Then you could use it like:

```rust
bold(|s| write!(s, "blah blah: {}", foo))?;
bold(|s| writeln!(s, "writing a line: {}", foo))?;
```

This saves creating the intermediary `String` with `format!` in the case that you need to format some coloured/bolded text.

The colours etc could be handled similarly? We should likely refactor `forc` to use a similar approach.

I think it might be OK to `unwrap` these results following the convention set by `println` (which `unwrap`s on I/O errors internally). Alternatively, the implementation could "fallback" to a normal `println` in the case that any of this custom formatted stream stuff produces an error.

Maybe a better name for this module could be `fmt` as it does more formatting than just colours (i.e. bold).

src/config.rs
```diff
@@ -0,0 +1,39 @@
...
+
+pub struct Config {
+    toolchains_dir: PathBuf,
+}
+
+impl Config {
+    pub(crate) fn from_env() -> Result<Self> {
+        let fuelup_dir = fuelup_dir();
+
+        let toolchains_dir = fuelup_dir.join("toolchains");
```
mitchmindtree, line 17: This could be a function (I think the `latest` toolchain PR might add one come to think of it).

src/config.rs
```diff
@@ -0,0 +1,39 @@
...
+    pub(crate) fn list_toolchains(&self) -> Result<Vec<String>> {
+        if self.toolchains_dir.is_dir() {
+            let toolchains: Vec<String> = fs::read_dir(&self.toolchains_dir)?
+                .filter_map(io::Result::ok)
+                .filter(|e| e.file_type().map(|f| f.is_dir()).unwrap_or(false))
+                .map(|e| {
+                    e.file_name()
+                        .into_string()
+                        .ok()
+                        .unwrap_or_else(|| "".to_string())
```
mitchmindtree, line 31:
```suggestion
                        .unwrap_or_default()
```

mitchmindtree (commented):
For the most part looking good!

Added a couple comments for now, but might be a little easier to review after landing #73 and merging or rebasing onto it.

src/colors.rs
binggh: Ah this is super useful - was wondering what the best way to do this would be, thanks for the tips!

At first was trying to see if `tracing` supports colored output but it turns out that the crate doesn't seem to support that and I had to fallback on writes/prints. In situations like [this](https://github.com/FuelLabs/fuelup/pull/64/files#diff-d9079e4f2b5f7f6151c0496aaf01876a092fb845fe2b7a67dd7f9f12d592d905R32-R33) it looks nicer to do `print_with_color()` and then `println!`, even though we already have `info!` from `tracing` - is there a better way to approach this?

src/fmt.rs
```diff
@@ -0,0 +1,23 @@
+use termcolor::{Color, ColorChoice, ColorSpec, StandardStream, WriteColor};
+
+pub fn bold<F>(write: F)
+where
+    F: FnOnce(&mut StandardStream) -> std::io::Result<()>,
+{
+    let mut stdout = StandardStream::stdout(ColorChoice::Auto);
+    stdout.set_color(ColorSpec::new().set_bold(true)).ok();
+    write(&mut stdout).expect("Unexpected error writing to stdout");
+    stdout.reset().ok();
```
mitchmindtree, line 10: A more conventional approach to ignoring `Result`s than `ok()` is to use `let _`, e.g.

```rust
let _ = stdout.set_color(ColorSpec::new().set_bold(true));
write(&mut stdout).expect("Unexpected error writing to stdout");
let _ = stdout.reset();
```

This avoids the conversion to `Option`, and I think is a generally renown approach to indicating that `Result`s are being ignored.

Also add a comment mentioning why we're ignoring these.

src/fmt.rs
```diff
@@ -0,0 +1,23 @@
...
+where
+    F: FnOnce(&mut StandardStream) -> std::io::Result<()>,
+{
+    let mut stdout = StandardStream::stdout(ColorChoice::Auto);
+    stdout
+        .set_color(ColorSpec::new().set_fg(Some(color)).set_bold(bold))
+        .ok();
+    write(&mut stdout).expect("Unexpected error writing to stdout");
+    stdout.reset().ok();
+}
```
mitchmindtree, line 23: Rather than using `bool` literals everywhere for `bold`, we should add dedicated functions for each behaviour, e.g. `colored` and `bold_colored`.

Also, in Rust, it's generally more conventional to have values precede functions within function arguments, e.g.

```rust
fn colored_bold<F>(color: Color, write: F)
```

This results in nicer formatting in the case that a large multi-line closure is provided, e.g.

```rust
colored_bold(Color::Green, |s| {
    writeln!(s, "blah blah blah: {}", foo);
    // more lines...
});
```
rather than
```rust
colored_bold(
    |s| {
        writeln!(s, "blah blah blah: {}", foo);
        // more lines...
    },
    Color::Green,
);

File: src/ops/fuelup_check.rs [hunk 6]
```diff
@@ -0,0 +1,146 @@
...
+    Ok(())
+}
+
+pub fn check(command: CheckCommand) -> Result<()> {
+    let CheckCommand { verbose } = command;
+
+    let cfg = Config::from_env()?;
+    let toolchains = cfg.list_toolchains()?;
+    let mut latest_versions: HashMap<String, Version> = HashMap::new();
+    let components = Vec::from([component::FORC, component::FUEL_CORE, component::FUELUP]);
```
mitchmindtree, line 61: When you want to construct a `Vec` from a fixed list, use the `vec!` macro:

```rust
let components = vec![copmonent::FORC, component::FUEL_CORE, component::FUELUP];
```

However remember `Vec`s are only necessary in the case that your list can grow. Instead, we should just use the fixed size array directly:

```rust
let components = [component::FORC, component::FUEL_CORE, component::FUELUP];

for &component in &components {
```

src/ops/fuelup_check.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    for component in components {
+        let download_cfg: DownloadCfg = DownloadCfg::new(component, None)?;
+        latest_versions.insert(component.to_string(), download_cfg.version);
+    }
+
+    for toolchain in toolchains {
+        if let Ok(toolchain) = Toolchain::from(&toolchain) {
```
mitchmindtree, line 69: Can this ever be an error? If we expect it to always be OK, we can `expect`, but if it can ever be an error, we should handle it.

src/ops/fuelup_check.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    match std::process::Command::new(component::FUELUP)
+        .arg("--version")
+        .output()
+    {
+        Ok(o) => {
+            let version = Version::parse(
+                String::from_utf8_lossy(&o.stdout)
+                    .split_whitespace()
+                    .collect::<Vec<&str>>()[1],
```
mitchmindtree, line 127: If we're only using the 2nd element, no need to collect the whole output into a `Vec`. It would be better to do something like:

```rust
let version = Version::parse(
    String::from_utf8_lossy(&o.stdout)
        .split_whitespace()
        .nth(1)
        .expect("expected version")
    )
);

File: src/ops/fuelup_check.rs [hunk 9]
```diff
@@ -0,0 +1,146 @@
...
+                if verbose && component == component::FORC {
+                    for plugin in SUPPORTED_PLUGINS {
+                        check_plugin(&toolchain, plugin, &latest_versions[component::FORC])?;
+                    }
+                }
+            }
+        }
+    }
+
+    match std::process::Command::new(component::FUELUP)
```
mitchmindtree, line 119: We shouldn't have to invoke the `fuelup` command. We **are** the fuelup command!

We should be able to get the version with something like this: https://docs.rs/clap/latest/clap/macro.crate_version.html

mitchmindtree (changes requested):

src/ops/fuelup_check.rs
binggh: Thanks for this catch. I used to store the `TargetTriple` within the `Toolchain` struct in one of the earlier commits and have since removed it - main reason being that custom toolchains that users create will probably not have a target triple out of convenience. Slipped my mind to remove this section that checks the Ok.

src/fmt.rs
binggh: Fixed with comment in [ad91ea4](https://github.com/FuelLabs/fuelup/pull/64/commits/ad91ea43ef7b20cdba817090bfb9ec3745f89f09) - is the comment clear?

mitchmindtree (approved):
:shipit:

PR #70, opened 2022-06-11
remove need_tty

binggh:
Addresses an issue in https://github.com/FuelLabs/sway-applications/pull/86

We do not actually use `need_tty` in fuelup since `fuelup` works slightly differently from `rustup`. 

In `rustup` the installer `rustup-init` is the one asking to prompt to install `rustup` and other components, but in our script at the moment we are installing `fuelup` directly and `fuelup` itself does not prompt users, and only act as a regular clap app.

After this is merged we should release v0.1.3 so that the sway-applications can use the updated curl script which supplies `--no-modify-path` to skip the prompt:

```sh
curl --proto '=https' --tlsv1.2 -sSf https://fuellabs.github.io/fuelup/fuelup-init.sh | sh -s -- --no-modify-path
```

### DIFF
--- a/fuelup-init.sh
+++ b/fuelup-init.sh
@@ -57,6 +57,7 @@ main() {
                     # don't attempt to interpret it.
                     continue
                 fi
+                ;;
         esac
     done
 
--- a/fuelup-init.sh
+++ b/fuelup-init.sh
@@ -42,8 +42,6 @@ main() {
         fi
     fi
 
-    # check if we have to use /dev/tty to prompt the user
-    local need_tty=yes
     # always prompt PATH modification, unless --no-modify-path provided
     local prompt_modify=yes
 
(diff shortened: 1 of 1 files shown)


Braqzen (approved):
I'm going to accept in order to unblock.
May want to wait for @mitchmindtree to review.

How come the previous implementation wasn't taking in `-- -y`?

binggh:
It was taking in ‘— -y’, but because all args were passed into ‘fuelup’, fuelup was the one which didn’t understand the command. The behaviour has since been changed though in master, we no longer pass all args to ‘fuelup’ and we call ‘install’ within the script itself (there are no other possible args for now anyway other than ‘install’!)

actually, if you are getting blocked on this and you want to test CI related stuff, you can curl from master instead of the published ‘fuelup-init.sh’ for now!

adlerjohn (approved):

PR #78, opened 2022-06-22
ensure dir exists when downloading components

binggh:
In some cases (especially when developing and running fuelup as `cargo run`), `fuelup` may not exist and thus the code doesn't know what to do with downloaded binaries (can't link since `fuelup` doesn't exist!) This is a small PR just to ensure that fuelup exists.

### DIFF
--- a/src/path.rs
+++ b/src/path.rs
@@ -1,4 +1,7 @@
-use std::path::PathBuf;
+use std::{
+    fs,
+    path::{Path, PathBuf},
+};
 
 use dirs;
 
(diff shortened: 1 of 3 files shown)


src/path.rs
```diff
@@ -23,3 +30,9 @@ pub fn toolchain_dir() -> PathBuf {
 pub fn toolchain_bin_dir(toolchain: &str) -> PathBuf {
     toolchain_dir().join(toolchain).join("bin")
 }
+
+pub fn ensure_dir_exists(path: &Path) {
+    if !path.is_dir() {
+        fs::create_dir_all(path).unwrap_or_else(|_| panic!("Failed to create {}", path.display()))
```
mitchmindtree, line 36: While unlikely, I'm not sure we want to `panic!` here - could we use `or_else` instead and propagate the result?

mitchmindtree (approved):

PR #84, opened 2022-06-23
correctly pass args to fuelup

binggh:
Fixes https://github.com/FuelLabs/action-fuel-toolchain/pull/30

Args to the executable have to be separated instead of being passed as a single, whitespace-separated string.


adlerjohn (commented):
Should there be a CI check for this? `Running fuelup-init.sh` in general I mean.

mitchmindtree (approved):

PR #92, opened 2022-06-26
fix bug with `index-versions.sh` that causes hashes to sometimes be empty

### LINKED ISSUE #91: bug with `index-versions.sh` that causes hashes to sometimes be empty
There is a small logic error with the way `index-versions.sh` was running that causes [empty hashes](https://github.com/FuelLabs/fuelup/blob/e513cec56660f0adc496fe7c16703e308e896111/channel-fuel-latest.toml#L9) to appear:

[`curl`](https://github.com/FuelLabs/fuelup/blob/a9847b7674bd252a996530cd099720e83083cf97/.github/workflows/scripts/index-versions.sh#L12) was not failing and exiting out of the script if the download was not successful and resulted in the script trying to calculate a hash for a missing file. As you can see from the CI workflow run below, there was a 404 but the workflow still completed successfully.

The solution is to exit out of the script should there be an attempt to curl a binary release that is not ready, and skip the indexing for that run.

https://github.com/FuelLabs/fuelup/runs/7048290707?check_suite_focus=true

bingcicle:
closes #91 

This PR fixes the issue of `index-versions` not properly handling missing binaries, and contains some minor refactoring.

- save the result of the `curl` so we can `grep` and exit if download was unsuccessful
- moved writing to `channel-fuel-latest.toml` to the end of the loop that processes all the downloading and hashing for each target of each binary, so that we do not attempt to write until all targets are successfully downloaded and hashed.

### DIFF
--- a/.github/workflows/scripts/index-versions.sh
+++ b/.github/workflows/scripts/index-versions.sh
@@ -9,10 +9,13 @@ latest_version() {
 add_url_and_hash() {
     _url="https://github.com/FuelLabs/$1/releases/download/v$2/$3"
     # shasum generates extra output so we take the first 64 bytes.
-    curl -sSf "${_url}" -L -o "${3}"
+    _err=$(curl -sSf "${_url}" -L -o "${3}" 2>&1)
+    if echo "$_err" | grep -q 404; then
+        printf "Could not download from %s - the release binary might not be ready yet. You can check if a binary is available here: https://github.com/FuelLabs/%s/releases/v%s\n" "${_url}" "$1" "$2"
+        exit 1
+    fi
     _hash=$(shasum -a 256 "$3" | head -c 64)
-    printf "url = \"%s\"\n" "${_url}" >>channel-fuel-latest.toml
-    printf "hash = \"%s\"\n\n" "${_hash}" >>channel-fuel-latest.toml
+    RETVAL="url = \"${_url}\"\nhash = \"${_hash}\"\n\n"
 }
 
 copy_same_pkg() {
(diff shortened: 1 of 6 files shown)


.github/workflows/scripts/index-versions.sh
```diff
@@ -48,11 +50,19 @@ create_pkg_in_channel() {
...
         for target in "${_targets[@]}"; do
-            printf "[pkg.%s.target.%s]\n" "$1" "${target}" >>channel-fuel-latest.toml
+            _content+="[pkg.${1}.target.${target}]\n"
             add_url_and_hash $_repo "$2" "$_tarball_prefix-${target}.tar.gz"
+            _content+="$RETVAL"
         done
+
+        # Only write to file if there's no problem downloading and hashing all the above releases.
+	_package=$(printf "%s\n%s" "${_header}" "${_content}")
+	echo -ne "$_package" >>channel-fuel-latest.toml
```
mitchmindtree, line 65: This indentation looks odd in the diff, but when looking at the raw file it [seems fine](https://github.com/FuelLabs/fuelup/blob/ec2d282f1f7ad8a1a6bdaa0fe8c26ec402ed9dbd/.github/workflows/scripts/index-versions.sh#L63-L65). :ghost: Just going to trust our bash fmt checker :)

mitchmindtree (dismissed):

.github/workflows/scripts/index-versions.sh
adlerjohn: This uses tabs instead of spaces. Should the formatter enforce that only spaces are used @bingcicle?

adlerjohn (approved):

.github/workflows/scripts/index-versions.sh
bingcicle: Manually ran `shfmt -i 4 -ci` for this particular file, I think I will open up another PR to enable `shellcheck` and `shfmt` for the `.github/workflows/scripts` directory since it's likely we add more scripts

mitchmindtree (approved):

.github/workflows/scripts/index-versions.sh
mitchmindtree: Nice one github, two different of rendering tab indentation on the same site :+1: :100:

PR #93, opened 2022-06-27
include `.github/workflows/scripts` dir in shfmt and shellcheck CI checks

bingcicle:
This includes both `index-versions.sh` and `verify-tags.sh` within the `/scripts` directory into `shellcheck` and `shfmt`, and fixes the issues with them.

**verify-tags.sh**
`SC2145`: https://www.shellcheck.net/wiki/SC2145
Also removed `cat $MANIFEST` since we can input `-f <FILE>`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,11 +21,11 @@ jobs:
         with:
           access_token: ${{ github.token }}
 
-  check-fuelup-init:
+  check-scripts:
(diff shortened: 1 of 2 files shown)


bingcicle:
There's this annoying thing with dangling checks after i change the names of the check `check-fuelup-init` -> `check-scripts` - not entirely sure what causes this 😕

PR #94, opened 2022-06-27
include .github/workflows/scripts dir in shfmt and shellcheck CI checks

bingcicle:
New PR in place of https://github.com/FuelLabs/fuelup/pull/93 since there was a CI error with a dangling `check-fuelup-init` check.

edit: seems like trying to re-open this didn't work - perhaps it doesn't matter since the old check is outdated anyway, just seems like an annoyance that happens when you edit step names.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,11 +21,11 @@ jobs:
         with:
           access_token: ${{ github.token }}
 
-  check-fuelup-init:
+  check-scripts:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


mitchmindtree (approved):

bingcicle:
Is there a way to ignore the `check-fuelup-init` check and force the merge? @mitchmindtree

mitchmindtree:
@bingcicle ahh I see! I think only @adlerjohn holds such power :mage_man:

adlerjohn:
It is done

PR #95, opened 2022-06-28
Test toolchain compatibility

### LINKED ISSUE #83: Add integration testing for version sets that must pass before updating version index
Related to #66, followup to #75.

I wonder if there's some way we can re-use the E2E tests from the Sway repo for this? E.g. fetch the sway repo, patch the `forc` and `fuel-core` versions used and run those tests?

We only really need the tests that check the integration of `forc` and `fuel-core`, so e.g. the contract deployment ones, running scripts and checking their results, etc. Perhaps we could filter out the language tests.

Alternatively, we can start adding our own integration tests here. That said, I can't think of many `forc`+`fuel-core` integration tests that would be only be beneficial here that would not *also* be beneficial at the Sway repo. It would be a shame to have some test here that could have been added to Sway and caught some incompatibility bug earlier in the release process.

bingcicle:
__WIP - Opening PR for visibility__

Closes #83 

__Note when reviewing:__ I've changed the workflow to trigger on `push` into this branch, instead of `schedule`. This is only for the purpose of testing the workflow for now, and it should be changed back to a scheduled cron job prior to merging.

_Please see details on the design within the comment [here](https://github.com/FuelLabs/fuelup/pull/95#issuecomment-1173928649)_

_Comment on how to review [here](https://github.com/FuelLabs/fuelup/pull/95#issuecomment-1177211625)_

### DIFF
--- a/.github/workflows/test-toolchain-compatibility.yml
+++ b/.github/workflows/test-toolchain-compatibility.yml
@@ -62,7 +62,7 @@ jobs:
           args: --manifest-path ./test/src/sdk-harness/Cargo.toml -- --test-threads=1 --nocapture
 
       # We have to explicitly check for:
-      # 1) e2e-tests.concluion == 'failure'
+      # 1) e2e-tests.conclusion == 'failure'
       # 2) std-lib-tests.conclusion == 'failure'
       # So that we create failure artifacts if and only if tests fail, and not the setup.
       - name: Create failure artifact
--- a/.github/workflows/test-toolchain-compatibility.yml
+++ b/.github/workflows/test-toolchain-compatibility.yml
@@ -121,21 +121,21 @@ jobs:
 
       # Run index-versions.sh here which generates the channel TOML file with download links and hashes for forc and fuel-core.
       - name: Prepare channel with compatible versions
+        if: ${{ env.LATEST_COMPATIBLE_FORC && env.LATEST_COMPATIBLE_FUEL_CORE }}
         run: |
             touch channel-fuel-latest.toml
             mkdir -p ${{ env.LATEST_CHANNEL_DIR }}
             ./.github/workflows/scripts/index-versions.sh ${{ env.LATEST_COMPATIBLE_FORC }} ${{ env.LATEST_COMPATIBLE_FUEL_CORE }}
             cp channel-fuel-latest.toml ${{ env.LATEST_CHANNEL_DIR }}
-        if: ${{ env.LATEST_COMPATIBLE_FORC && env.LATEST_COMPATIBLE_FUEL_CORE }}
 
       - name: Deploy latest channel
+        if: ${{ env.LATEST_COMPATIBLE_FORC && env.LATEST_COMPATIBLE_FUEL_CORE }}
         uses: peaceiris/actions-gh-pages@v3
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: ${{ env.LATEST_CHANNEL_DIR }}
           keep_files: true
           destination_dir: ./
-        if: ${{ env.LATEST_COMPATIBLE_FORC && env.LATEST_COMPATIBLE_FUEL_CORE }}
 
       - name: Deploy incompatible versions
         uses: peaceiris/actions-gh-pages@v3
--- a/.github/workflows/test-toolchain-compatibility.yml
+++ b/.github/workflows/test-toolchain-compatibility.yml
@@ -38,7 +38,7 @@ jobs:
 
   test-toolchain-compatibility:
     needs: setup-tests
-    if: ${{ env.forc-${{ matrix.forc-version }}@fuel-core-${{ matrix.fuel-core-version }} == true }}
+    if: ${{ "${{ env.forc-${{ matrix.forc-version }}@fuel-core-${{ matrix.fuel-core-version }} }}" == true }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false 
@@ -51,6 +51,7 @@ jobs:
         ports:
           - 4000:4000
 
+    steps:
       - name: Checkout Sway repo
         uses: actions/checkout@v3
         with:
--- a/.github/workflows/test-toolchain-compatibility.yml
+++ b/.github/workflows/test-toolchain-compatibility.yml
@@ -1,6 +1,14 @@
 name: Test toolchain compatibility (Latest)
 
 on:
+  workflow_dispatch:
+    inputs:
+      forc-versions:
+        required: true
+        type: string
+      fuel-core-versions:
+        required: true
+        type: string
   workflow_call:
     inputs:
       forc-versions:
--- a/.github/workflows/test-toolchain-compatibility.yml
+++ b/.github/workflows/test-toolchain-compatibility.yml
@@ -68,10 +68,13 @@ jobs:
           command: test
           args: --manifest-path ./test/src/sdk-harness/Cargo.toml -- --test-threads=1 --nocapture
 
+      # We have to explicitly check for:
+      # 1) e2e-tests.concluion == 'failure'
+      # 2) std-lib-tests.conclusion == 'failure'
+      # So that we create failure artifacts if and only if tests fail, and not the setup.
       - name: Create failure artifact
         if: ${{ failure() && (steps.e2e-tests.conclusion == 'failure' || steps.std-lib-tests.conclusion == 'failure' )}}
-        run: |
-          touch incompatible-forc-${{ matrix.forc-version }}@fuel-core-${{ matrix.fuel-core-version }}
+        run: touch incompatible-forc-${{ matrix.forc-version }}@fuel-core-${{ matrix.fuel-core-version }}
 
       - name: Upload failure artifact
         if: ${{ failure() && (steps.e2e-tests.conclusion == 'failure' || steps.std-lib-tests.conclusion == 'failure' ) }}
@@ -81,10 +84,10 @@ jobs:
           name: incompatible-forc-${{ matrix.forc-version }}@fuel-core-${{ matrix.fuel-core-version }}
           path: incompatible-forc-${{ matrix.forc-version }}@fuel-core-${{ matrix.fuel-core-version }}
 
+      # Success artifacts are created if everything above passed. This marks the 2 versions as compatible.
       - name: Create success artifact
         if: ${{ success() }}
-        run: |
-          touch compatible-forc-${{ matrix.forc-version }}@fuel-core-${{ matrix.fuel-core-version }}
+        run: touch compatible-forc-${{ matrix.forc-version }}@fuel-core-${{ matrix.fuel-core-version }}
 
       - name: Upload success artifact
         if: ${{ success() }}
@@ -95,7 +98,7 @@ jobs:
           path: compatible-forc-${{ matrix.forc-version }}@fuel-core-${{ matrix.fuel-core-version }}
 
   index-versions:
-    name: Log latest compatible versions that passed tests
+    name: Index versions compatibilities
     needs: test-toolchain-compatibility
     if: needs.test-toolchain-compatibility.result != 'skipped'
     runs-on: ubuntu-latest
@@ -106,6 +109,9 @@ jobs:
         with:
           path: versions
 
+      # Here, we decide which versions to mark as compatible or incompatible.
+      # Compatible: the uploaded artifact starting with the name 'compatible*' and is at the top when sorted.
+      # Incompatible: all the uploaded artifacts that has 'incompatible' in file names, followed by their versions.
       - name: Prepare compatible and incompatible versions
         id: prep-versions
         working-directory: versions
@@ -120,12 +126,14 @@ jobs:
               touch ${{ env.INCOMPATIBLE_DIR }}/$version
             done
 
+      # Run index-versions.sh here which generates the channel TOML file with download links and hashes for forc and fuel-core.
       - name: Prepare channel with compatible versions
         run: |
             touch channel-fuel-latest.toml
             mkdir -p ${{ env.LATEST_CHANNEL_DIR }}
             ./.github/workflows/scripts/index-versions.sh ${{ env.LATEST_COMPATIBLE_FORC }} ${{ env.LATEST_COMPATIBLE_FUEL_CORE }}
             cp channel-fuel-latest.toml ${{ env.LATEST_CHANNEL_DIR }}
+        if: ${{ env.LATEST_COMPATIBLE_FORC && env.LATEST_COMPATIBLE_FUEL_CORE }}
 
       - name: Deploy latest channel
         uses: peaceiris/actions-gh-pages@v3
--- a/.github/workflows/test-toolchain-compatibility.yml
+++ b/.github/workflows/test-toolchain-compatibility.yml
@@ -89,7 +89,7 @@ jobs:
   index-versions:
     name: Log latest compatible versions that passed tests
     needs: test-toolchain-compatibility
-    if: ${{ needs.test-toolchain-compatibility.result != "skipped" }}
+    if: needs.test-toolchain-compatibility.result != "skipped"
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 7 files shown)


bingcicle:
Hm, it just occurred to me that a scenario may occur here that _might_ affect how this works. If there are fast succession of releases of `fuel-core` or `forc` (for example, if we cut 2 separate releases in the time between 2 runs of this workflow), this method of indexing might fail, since we would miss the earlier version.

bingcicle:
I think I will re-write the comparison of versions in Rust in a short script - the solution to the above would be to simply, instead of assuming that there's only 1 latest version every 30 minutes, just collect the versions (for both `forc` and `fuel-core` from the last published within `channel-fuel-latest.toml`, and execute workflows for those jobs.

adlerjohn:
> Although I'm not sure if it would ever be realistic that we cut 2 releases in quick succession as described above.

Actually, we've done exactly that many times in the past 😂

bingcicle:
This is almost ready. Took a while longer than expected since it's pretty difficult to test out actions and get quick feedback to test out the flow. Have a personal repo at https://github.com/bingcicle/index-versions that you can check out to see how the workflow works.

### To be done:

- [ ] compare-version actual script
- [ ] include actual tests within workflow

## Components:

### compare-versions

Rust-based script, simply parses the latest channel in gh-pages, queries the [workflow runs API](https://docs.github.com/en/rest/actions/workflow-runs#list-workflow-runs-for-a-repository) collects and prints the range of new `forc` and `fuel-core` versions between the latest published version on the channel and the latest released version, provided they are not found to be **incompatible** (explained later).

The printing is done so we can use them as inputs to `test-toolchain-compatibility.yml` workflow later. This will be in a json array format.

If there were new releases in one repo detected but no new release for the other repo (sway/fuel-core or vice versa), we will print the last published version for the other repo that is known to be stable.

If there were new releases for **both** repos, we have to print the new versions + the last published versions for both repos, since we do not know if the latest releases will pass with each other.

Currently the script is only fake printing versions for my convenience in testing the workflow in my personal repo.

### index-versions.yml

This is the entrypoint of the workflow and is run every 30 minutes - we first run `compare-versions`, and if it doesn't print any new versions, we set `should_run` to false and do not enter the next step. The workflow will then end here and no unnecessary CI time is consumed.

### test-toolchain-compatibility.yml

This is the meat of the workflow, if `compare-versions` prints out new versions for either `forc` or `fuel-core`, we will take them in and use `matrix` to run tests in parallel for `forc` by checking out the Sway repo at the given version(s) and pulling `fuel-core` the Docker image(s) at the given version(s). (we need to do this since we cannot assume that there will be only 1 new release between runs).

_(I'm not sure if the above would be over-complicating things but I couldn't think of another way to ensure that multiple releases in quick succession can be covered without doing it this way. Releases in quick succession could happen as an edge case, even if it is most probable that there will only be 1 release between 2 runs of this CI._

As mentioned above we will run tests which use `fuel-core` as a service in the Sway CI [here](https://github.com/FuelLabs/sway/blob/823707d5939ad6112bfc5a0a538d58b4d7d14a99/.github/workflows/ci.yml#L229-L270). For convenience's sake I am also not running tests in my personal repo but merely exiting with 1 if `fuel-core` version is `0.9.4` to simulate tests failing.

### Outcomes of tests

1) If tests pass here, we mark versions as **compatible** and **incompatible** otherwise: [example](https://github.com/bingcicle/index-versions/tree/423e0d0de7876c77c398f0acf25e5d78b7880a86)

This is a simulation of `fuel-core-0.9.4` being released and failing tests with `forc` versions. `channel-latest-fuel.toml` is **not** updated, and the new versions are marked as **incompatible** within the dir `incompatible-versions`

2) Now `forc-0.17.0` is released, and we test it against `fuel-core-0.9.4` and it fails again: [example](https://github.com/bingcicle/index-versions/tree/f11e09cf87c180a7953258e01d64ea212a7302de). You can see that the channel was not updated, but a new file was added to `incompatible-versions` which marks the versions as incompatible. We do not test it again at the next interval, since it is already said to be incompatible.

Versions are only deemed **incompatible** only if **tests** fail, since the setup could fail for other reasons. This could be docker issues or github api failure, or other reasons unrelated to tests failing.

I simulated it [here](https://github.com/bingcicle/index-versions/runs/7177619605?check_suite_focus=true) with a non-existent forc version. If the setup fails, we should fail the CI but not mark the versions as incompatible and retry it at the next interval. Notice we skip creating new failure/success files here.

### Ending notes

I think in a private conversation with @mitchmindtree, we discussed the possibility naming the latest compatible versions the `stable` toolchain since the versions would technically pass all tests under our CI already. We can label the latest of each release, regardless of whether they pass tests, as the true `latest`.

I think 1 concern I have with this approach is the recording of incompatibilities - it is indeed convenient to publish them to gh-pages but I wonder if there's a better alternative so that we don't end up accidentally deleting the entire directory or even the branch, since testing compatibility relies on tests (which is expensive to run).

adlerjohn:
> I wonder if there's a better alternative so that we don't end up accidentally deleting the entire directory or even the branch

We can always add branch protection rules to prevent deletion and force pushes. That will ensure at least that branch history is not lost and can be restored if it gets corrupted.

bingcicle:
I think this should be good to go for a review. 

Not sure how to go about reviewing this if I were asked for a review since I basically completely wrecked the commit history, but I will list some things to review. Ultimately, the goal should be to have a workflow that runs every set interval to check for new releases, and update our official distribution channel `channel-fuel-latest.toml` in the `fuelup` gh-pages.

## compare-versions

This is a Rust script which get a range of new versions since the channel was last updated, and should provide a list of strings in the format `forc-<FORC_VERSION>@fuel-core-<FUEL_CORE_VERSION>` (eg. `forc-0.17.0@fuel-core-0.16.2`). For example, The [current published one](https://github.com/FuelLabs/fuelup/blob/d2cb7b6c90d74f89c63abc4891818ed6ec5cf8b4/channel-fuel-latest.toml) has `forc` at `0.17.0` - if subsequently `0.17.1` and `0.17.2` are released, these versions should be picked up by this script and matched with the latest `fuel-core` (currently `0.9.4` to form the formatted strings. I wrote some basic tests to show this behaviour.

## index-versions.yml

The main step to review here is the **Compare versions and set variables for tests if needed** step.

This should parse the output of `compare-versions` above and produce a matrix of `forc` and `fuel-core` versions for testing. It should also check if there are incompatibilities published onto gh-pages based on the formatted string.

This should not initiate tests if 1) there are no new versions detected by `compare-versions`, or 2) the versions detected by `compare-versions` exists as file stubs within `gh-pages/incompatible-versions` directory. [Example](https://github.com/bingcicle/index-versions-test/tree/75cfc970a652924ed4e96827099ac6fc2bb9979a)

## test-toolchain-compatibility.yml

This workflow checks out the Sway repo at the above given forc version as well as pulls the `fuel-core` image at the above given `fuel-core` version.

The main logic to check here is the flow from testing to success/failure actions. In an event of a success, we create file stubs in the format `compatible-forc-<FORC_VERSION>@fuel-core-<FUEL_CORE_VERSION>`. In an event of a failure, we create file stubs in the format `incompatible-forc-<FORC_VERSION>@fuel-core-<FUEL_CORE_VERSION>`. We upload these file stubs using `upload-artifacts` action.

Note that I specifically check for test failures in creating the incompatible files - we do not want the workflow to mark versions as incompatible if setup fails, eg. if we failed to pull the `fuel-core` image due to ghcr being down.

If the `test-toolchain-compatibility` job is not skipped, we run the `index-versions` job. This job should:

1) `ls` all the files starting with `compatible-*` and take the latest forc and fuel-core versions, run the `index-versions.sh` script from the previous PR to create `channel-fuel-latest.toml`. The current channel-fuel-latest.toml` is a result of a successful run of this, [here](https://github.com/FuelLabs/fuelup/blob/d2cb7b6c90d74f89c63abc4891818ed6ec5cf8b4/channel-fuel-latest.toml)

2) `ls` all the files starting with `incompatible-*` and publish them as files on gh-pages so that we don't re-run the tests based on the same versions.

## Reusability
The workflows should also be reviewed for reusability, ultimately we might want to do the same for `nightly` and `stable` (unless this is our stable?). I separated out the `test-toolchain-compatibility` workflow for this reason, since we would be able to have a different workflow to trigger these tests on a different commit of `forc` and/or `fuel-core`.

scripts/compare-versions/src/main.rs
```diff
@@ -0,0 +1,228 @@
+use anyhow::Result;
```
mitchmindtree, line 1: Perhaps  we could at some crate root documentation here (e.g. `//!`) that describes the purpose of this script. E.g. that the purpose is to fetch the recent workflow runs to find new versions and then outputs all combinations of untested versions to `stdout`

scripts/compare-versions/src/main.rs
```diff
@@ -0,0 +1,228 @@
...
+
+    select_versions(&channel_doc, forc_versions, fuel_core_versions);
+    Ok(())
+}
+
+fn select_versions(
+    channel: &Document,
+    mut forc_versions: Vec<Version>,
+    mut fuel_core_versions: Vec<Version>,
+) -> String {
```
mitchmindtree, line 154: This function actually performs the I/O and prints the main output of the program, but it's a little unclear looking at the signature. Perhaps rather than calling `print_selected_versions` within this function, we could instead move that into a step *after* `select_versions` is called?

mitchmindtree (commented):
Thanks again for the chat and for walking me through all this! Definitely learnt a lot about GitHub actions :octocat: 

As mentioned in our chat, it'd be great to have a `CHANNELS.md` document that details how each channel's versions are selected, tested and updated (just with latest for now). I can imagine something like this helping a lot both for future maintenance, as well as for curious users who want to know exactly what "latest" or "nightly" means.

scripts/compare-versions/src/main.rs
```diff
@@ -0,0 +1,228 @@
...
+    let latest_forc_indexed_version = parse_latest_indexed_version(channel, "forc");
+    let latest_fuel_core_indexed_version = parse_latest_indexed_version(channel, "fuel-core");
+
+    match (forc_versions.is_empty(), fuel_core_versions.is_empty()) {
+        (true, false) => forc_versions.push(latest_forc_indexed_version),
+        (false, true) => fuel_core_versions.push(latest_fuel_core_indexed_version),
+        (false, false) => {
```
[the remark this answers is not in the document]
bingcicle: Actually, the intent is that if there are 2 new released versions, testing them against each other might fail. In this scenario, we want to pull the last published versions as well, because it's possible that one or both of the 2 released versions pass against the last published `forc` or `fuel-core`, and we want to update the channel if this happens.

> If the only combo of versions we are left with comes from the existing index toml which has already been tested successfully in the past, then we already know those versions are successfully compatible so there should be no need to test them again right?

Yep, hence if both `forc_versions` and `fuel_core_versions` are empty, we don't print anything.

CHANNELS.md
```diff
@@ -0,0 +1,46 @@
...
+
+The `latest` channel is possibly updated by a scheduled GitHub workflow that **runs every 30 minutes** and checks for new releases of `forc` and `fuel-core`.
+
+### How the CI updates the `latest` channel
+
+A channel is automatically updated if integration tests pass for the specific version of `forc` and `fuel-core`. This channel is typically found in a TOML file, and for the `latest` channel, is named `channel-fuel-latest.toml`.
+
+### How the workflow works, a summary
+
+The short explanation of how the scheduled workflow works is that we fetch all the latest versions after the latest published versions within `channel-fuel-latest.toml`, and we run tests for the possible combinations of them if not already marked as incompatible. After tests are done we either mark the version sets as incompatible if the tests failed or update the channel.
```
mitchmindtree, line 13: These "How the CI updates the `latest` channel" and "How the workflow works, a summary" sections sound a little rambly and could possibly be structured a little better.

How about something along these lines:


> # Channels
> 
> *Introduce the concept of channels...*
> 
> ### `latest`
> 
> The `latest` channel is `fuelup`'s default channel. It provides access to the latest compatible, published releases of `forc` and `fuel-core`.
> 
> When installing the `latest` channel, `fuelup` will refer to the `channel-fuel-latest.toml` file within this repository to determine the set of `forc` and `fuel-core` versions to retrieve. The versions in this file are updated by a scheduled GitHub workflow that runs once every 30 minutes and performs the following steps:
> 
> 1. Checks for newly published versions of `forc` and `fuel-core`.
> 2. Tests compatibility of new versions against a set of integration tests.
> 3. Selects the latest set of versions that successfully pass the tests.
> 4. Publishes the selected versions to the `channel-fuel-latest.toml` manifest.
> 
> 
> 
> ### `nightly`
> 
> *TODO*
>
> ---

CHANNELS.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Developer Guide
+
+### Understanding the workflow
```
mitchmindtree, line 17:
```suggestion
### Understanding the `latest` workflow
```

mitchmindtree: We might want to have different sections for latest and nightly, as the workflows will likely be a bit different?

mitchmindtree (commented):
Nice, I think these new comments and docs will be a huge help for future contributors/maintainers (including us)!

I think some of the writing under the new `CHANNELS.md` could be a little more concise, but otherwise they're a big improvement over having to work everything out by reading through the workflow itself.

.github/workflows/test-toolchain-compatibility.yml
```diff
@@ -61,10 +68,18 @@ jobs:
...
+      # See:
+      # https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts
+      #
+      # On failure, we upload files with the `incompatible` prefix.
+      #
       # We have to explicitly check for:
       # 1) e2e-tests.conclusion == 'failure'
       # 2) std-lib-tests.conclusion == 'failure'
-      # So that we create failure artifacts if and only if tests fail, and not the setup.
+      # So that we create failure artifacts iff tests fail and not anywhere before so that we can retry this workflow
```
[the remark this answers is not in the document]
bingcicle: `iif` is `if and only if` but i guess it's not appropriate to use it here 😅

CHANNELS.md
```diff
@@ -0,0 +1,46 @@
+# Channels
+
+`fuelup` adopts a simplified version of `rustup` [channels](https://rust-lang.github.io/rustup/concepts/channels.html). Currently, only the `latest` channel will be published and serve as a source of distribution of Fuel toolchain binaries.
+
+The `latest` channel is possibly updated by a scheduled GitHub workflow that **runs every 30 minutes** and checks for new releases of `forc` and `fuel-core`.
```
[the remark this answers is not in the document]
bingcicle: hmm, I used `possibly` because I wanted to say that the scheduled workflow doesn't always update the channel. Not sure if there's a better way to put this point across

CHANNELS.md
bingcicle: I added the `nightly` section with a bit more explanation that we'll support it soon.

mitchmindtree (commented):
OK, I think this is good to go!

@bingcicle I think we just need to change our push trigger back to the schedule one, and that's it right?

Also, your comment [here](https://github.com/FuelLabs/fuelup/pull/103#issuecomment-1182953720) mentions some flakeyness in tests here? Is your preference to land #103 first or this?

bingcicle:
> OK, I think this is good to go!
> 
> @bingcicle I think we just need to change our push trigger back to the schedule one, and that's it right?

Yup!

> Also, your comment [here](https://github.com/FuelLabs/fuelup/pull/103#issuecomment-1182953720) mentions some flakeyness in tests here? Is your preference to land #103 first or this?

The PR in #103 is meant to fix fetching the latest version. The flakiness was a result of the GitHub API not returning anything at certain times. I had to rerun the CI here twice to make running `fuelup-init` pass because the version fetched was an empty string and I'm not entirely sure why that happens. I think some contributors also ran into this problem, and running it after a while seemed to work. 

I'm guessing it's an issue with the API or the way that the script was querying the API, but either way querying the `fuelup-version` file in the changes within #103 is better than relying on the API at all times imo.

mitchmindtree (approved):

PR #97, opened 2022-06-29
Fix fuelup self update

### LINKED ISSUE #96: `fuelup self update` fails with ETXTBSY
This the output I get when running `fuelup self update`
```
Fetching binary from https://github.com/FuelLabs/fuelup/releases/download/v0.2.1/fuelup-0.2.1-x86_64-unknown-linux-gnu.tar.gz
Unpacking and moving fuelup to /home/segfault_magnet/.fuelup/bin
fuelup failed to update itself: Text file busy (os error 26)
```

If I run it with `strace` I get:
```
openat(AT_FDCWD, "/home/segfault_magnet/.fuelup/bin/fuelup", O_WRONLY|O_CREAT|O_TRUNC|O_CLOEXEC, 0100755) = -1 ETXTBSY (Text file busy)
close(5)                                = 0
close(4)                                = 0
close(3)                                = 0
write(1, "fuelup failed to update itself: "..., 61fuelup failed to update itself: Text file busy (os error 26)
) = 61
```

This [SO question](https://stackoverflow.com/questions/1712033/replacing-a-running-executable-in-linux) might be useful

bingcicle:
Fixes #96

`fuelup` should remove itself (unlink) before trying to download a new version. ~We will temporarily copy this as a backup in case download fails - in which case we copy the temp `fuelup` back.~

Updated with @mitchmindtree 's suggestion:

1. Create a tempdir `fuelup_new_dir` and fetch new fuelup to this dir. At this point, if download fails, do nothing else so old fuelup remains.
2. Upon success, if old fuelup already exists, rename old fuelup to `fuelup-backup`.
3. Finally, copy the new fuelup within `fuelup_new_dir` into the `bin` directory.
4. If the above step somehow fails, rename `fuelup-backup` back into the `fuelup`. If this also somehow fails, we recommend to reinstall via the `fuelup-init` script.
5. We do not need to clean up the download since we're using [`tempdir`](https://docs.rs/tempfile/3.3.0/tempfile/fn.tempdir.html) which cleans itself up once we exit the scope.

### DIFF
--- a/src/ops/fuelup_self.rs
+++ b/src/ops/fuelup_self.rs
@@ -24,8 +24,9 @@ pub fn self_update() -> Result<()> {
     let fuelup_backup = fuelup_bin_dir.join("fuelup-backup");
 
     let fuelup_new_dir = tempdir_in(&fuelup_bin_dir)?;
+    let fuelup_new_dir_path = fuelup_new_dir.path();
 
-    if let Err(e) = attempt_install_self(download_cfg, fuelup_new_dir.path()) {
+    if let Err(e) = attempt_install_self(download_cfg, fuelup_new_dir_path) {
         // Skip all other steps if downloading fails here.
         // We do not need to handle failure, since this downloads to a tempdir.
         bail!("Failed to install fuelup: {}", e);
@@ -38,10 +39,10 @@ pub fn self_update() -> Result<()> {
 
     info!(
         "Moving {} to {}",
-        fuelup_new_dir.path().join("fuelup").display(),
+        fuelup_new_dir_path.join("fuelup").display(),
         &fuelup_bin.display()
     );
-    if let Err(e) = fs::rename(fuelup_new_dir.path().join("fuelup"), &fuelup_bin) {
+    if let Err(e) = fs::rename(fuelup_new_dir_path.join("fuelup"), &fuelup_bin) {
         error!("Failed to replace the old fuelup: {}", e);
         error!("Restoring old fuelup");
 
--- a/src/ops/fuelup_self.rs
+++ b/src/ops/fuelup_self.rs
@@ -1,4 +1,4 @@
-use anyhow::{anyhow, bail, Result};
+use anyhow::{bail, Result};
 use std::{fs, path::Path};
 use tempfile::tempdir_in;
 use tracing::{error, info};
--- a/src/ops/fuelup_self.rs
+++ b/src/ops/fuelup_self.rs
@@ -50,13 +50,13 @@ pub fn self_update() -> Result<()> {
         fuelup_new.display(),
         &fuelup_bin.display()
     );
-    if let Err(e) = fs::rename("wot", &fuelup_bin) {
+    if let Err(e) = fs::rename(&fuelup_new, &fuelup_bin) {
         error!(
             "Failed to replace old fuelup with new fuelup. Attempting to restore backup fuelup.",
         );
         // If we have failed to replace the old fuelup for whatever reason, we want the backup.
         // Although unlikely, should this last step fail, we will recommend to re-install fuelup using fuelup-init.
-        if let Err(e) = fs::rename(&"wot", &fuelup_bin) {
+        if let Err(e) = fs::rename(&fuelup_backup, &fuelup_bin) {
             bail!(
                 "Could not restore backup fuelup. {}
 
--- a/src/ops/fuelup_self.rs
+++ b/src/ops/fuelup_self.rs
@@ -22,16 +22,24 @@ pub fn self_update() -> Result<()> {
     let fuelup_backup = temp_dir().join("fuelup-backup");
 
     if fuelup_bin.exists() {
-        // Make a backup of fuelup, in case downloading fails. If download fails below, we use the backup.
+        // Make a backup of fuelup.
         fs::copy(&fuelup_bin, &fuelup_backup).expect("Could not make a fuelup backup");
         // We cannot copy new fuelup over the old; we must unlink it first.
         fs::remove_file(&fuelup_bin).expect("Could not remove fuelup");
     };
 
     if let Err(e) = attempt_install_self(download_cfg, &fuelup_bin_dir()) {
-        error!("Failed to install and replace fuelup - {}.", e);
-        fs::copy(fuelup_backup, fuelup_bin)?;
+        error!("Failed to install and replace fuelup. {}", e);
+
+        // We want to restore the backup fuelup in case download goes wrong.
+        if fuelup_backup.exists() {
+            fs::copy(&fuelup_backup, &fuelup_bin).expect("Could not restore fuelup-backup");
+        }
     };
+
+    if fuelup_backup.exists() {
+        fs::remove_file(&fuelup_backup).expect("Could not remove fuelup backup");
+    }
 
     Ok(())
 }
--- a/src/ops/fuelup_self.rs
+++ b/src/ops/fuelup_self.rs
@@ -1,4 +1,4 @@
-use anyhow::{bail, Result};
+use anyhow::{anyhow, bail, Result};
 use std::{fs, path::Path};
 use tempfile::tempdir_in;
 use tracing::{error, info};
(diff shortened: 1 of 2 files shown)


src/ops/fuelup_self.rs
```diff
@@ -1,17 +1,43 @@
...
+    };
+
+    if let Err(e) = attempt_install_self(download_cfg, &fuelup_bin_dir()) {
+        error!("Failed to install and replace fuelup. {}", e);
 
-    download_file_and_unpack(&download_cfg, &fuelup_bin_dir)?;
-    unpack_bins(&fuelup_bin_dir, &fuelup_bin_dir)?;
+        // We want to restore the backup fuelup in case download goes wrong.
+        if fuelup_backup.exists() {
+            fs::copy(&fuelup_backup, &fuelup_bin).expect("Could not restore fuelup-backup");
```
mitchmindtree, line 38: While this error path should be rare, we should return an `Err` that also recommends reinstalling `fuelup` as if this fails we're left with no `fuelup` bin on `PATH`.

mitchmindtree: Upon success, we might want to cleanup the `fuelup_backup` too just because we can:

```rust
// Ignore the error case - this was located in a `temp` directory anyway.
let _ = fs::remove_file(&fuelup_backup);
```

mitchmindtree (commented):
Currently, it looks like this PR takes the following approach:

1. Move existing `fuelup` bin to a temp directory as `fuelup-backup`.
2. Attempt to install new `fuelup`.
3. Upon failure, copy `fuelup-backup` back to original directory.

Generally when replacing an important file,  it's better to not touch the original until we have successfully fetched/created the new one in order to reduce the number of possible I/O error cases that might occur.

Perhaps we should consider doing something like:

1. Fetch new `fuelup` as `fuelup-new` to bin directory.
2. Upon success, rename old `fuelup` to `fuelup-backup` in same directory.
3. Rename `fuelup-new` to `fuelup`.
4. Remove `fuelup-backup`.

This means that the only operation that might fail that could leave us without a valid `fuelup` is the rename from `fuelup-new` to `fuelup`, which is perhaps less likely to fail than a `copy`. It also reduces the window of time that the `fuelup-backup` might be corrupted as rather than waiting for the entirety of the `fuelup` download, it should only exist for the instant that the renames occur.

Thoughts?

bingcicle:
> 1. Fetch new `fuelup` as `fuelup-new` to bin directory.
> 2. Upon success, rename old `fuelup` to `fuelup-backup` in same directory.
> 3. Rename `fuelup-new` to `fuelup`.
> 4. Remove `fuelup-backup`.


Ah yeah this approach sounds better - will update the behaviour.

By the way, I converted this back to a draft because I'm still running into the `Text file busy` error despite the current implementation (Testing it on a linode Linux instance)

src/ops/fuelup_self.rs
```diff
@@ -1,17 +1,43 @@
+use std::{fs, path::Path};
+use tempfile::tempdir;
+use tracing::error;
+
```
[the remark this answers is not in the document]
bingcicle: Not quite sure what you mean by this?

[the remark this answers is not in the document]
bingcicle: Do you mean that it should all be in lines without line breaks?

bingcicle:
Misread/misunderstood your original comment - have replaced `fs::copy` with `fs::rename` where possible.

Did not do so for downloading new fuelup version, since that would require refactoring the download module which could affect `fuelup toolchain install latest` because the `unpack` fn moves `fuelup` binary out of the containing folder as is without renaming it.

src/ops/fuelup_self.rs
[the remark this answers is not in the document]
mitchmindtree: Yes I think the general convention is to have no line breaks between all top-level use statements (including the `crate` one below). I believe after removing the line breaks, rustfmt will then also order the use statements alphabetically. Doing so makes it a bit easier for a reader to kind of "binary search" the use statements if they're looking to edit a particular set of imports for example.

src/ops/fuelup_self.rs
```diff
@@ -1,17 +1,60 @@
...
+    if let Err(e) = attempt_install_self(download_cfg, fuelup_new_dir.path()) {
+        // Skip all other steps if downloading fails here.
+        // We do not need to handle failure, since this downloads to a tempdir.
+        bail!("Failed to install fuelup: {}", e);
+    };
+
+    if fuelup_bin.exists() {
+        // Make a backup of fuelup, fuelup-backup.
+        fs::rename(&fuelup_bin, &fuelup_backup).expect("Could not make a fuelup-backup");
```
mitchmindtree, line 36: We should return an error here

src/ops/fuelup_self.rs
```diff
@@ -1,17 +1,60 @@
...
+    };
+
+    info!(
+        "Copying {} to {}",
+        fuelup_new_dir.path().join("fuelup").display(),
```
mitchmindtree, line 41: We should construct this path once and re-use it here and at L44, especially as the name contains a string literal. Do we have this name as a const anywhere?

src/ops/fuelup_self.rs
```diff
@@ -1,17 +1,60 @@
...
+        bail!("Failed to install fuelup: {}", e);
+    };
+
+    if fuelup_bin.exists() {
+        // Make a backup of fuelup, fuelup-backup.
+        fs::rename(&fuelup_bin, &fuelup_backup).expect("Could not make a fuelup-backup");
+    };
+
+    info!(
+        "Copying {} to {}",
```
mitchmindtree, line 40:
```suggestion
        "Moving {} to {}",
```

src/ops/fuelup_self.rs
```diff
@@ -1,17 +1,60 @@
...
+    if let Err(e) = fs::rename(fuelup_new_dir.path().join("fuelup"), &fuelup_bin) {
+        error!("Failed to replace the old fuelup: {}", e);
+
+        // If we have failed to replace the old fuelup for whatever reason, we want the backup.
+        // Should this last step fail, we will recommend to re-install fuelup using fuelup-init.
+        if let Err(e) = fs::rename(&fuelup_backup, &fuelup_bin) {
+            error!("Could not restore backup fuelup: {}", e);
+            error!("You should re-install fuelup using fuelup-init:");
+            error!("`curl --proto '=https' --tlsv1.2 -sSf https://fuellabs.github.io/fuelup/fuelup-init.sh | sh`");
+        }
```
mitchmindtree, line 53: After this, it looks like we should return an error? Otherwise it seems the program will exist with a "success" exit code indicating that nothing went wrong.

mitchmindtree (changes requested):

src/ops/fuelup_self.rs
bingcicle: Hmm, we don't have it as a const and I didn't include this within `path.rs` as I believe this is only relevant within the context of `fuelup self update`. `fuelup_new_dir` is also a `TempDir` (which I opted to use since it automatically cleans up after itself). I will create the path right after execution of `tempdir_in()`.

src/ops/fuelup_self.rs
```diff
@@ -1,17 +1,70 @@
...
+        error!("Restoring old fuelup");
+
+        // If we have failed to replace the old fuelup for whatever reason, we want the backup.
+        // Although unlikely, should this last step fail, we will recommend to re-install fuelup using fuelup-init.
+        if let Err(e) = fs::rename(&fuelup_backup, &fuelup_bin) {
+            error!("Could not restore backup fuelup: {}", e);
+            error!("You should re-install fuelup using fuelup-init:");
+            error!("`curl --proto '=https' --tlsv1.2 -sSf https://fuellabs.github.io/fuelup/fuelup-init.sh | sh`");
+            std::process::exit(1);
+        }
```
mitchmindtree, line 63: A couple things here:

1. We should avoid adding a `std::process::exit` in the middle of an unsuspecting function like `self_update`. `std::process::exit` terminates the program and does not return, which might be confusing for a future maintainer if we want to refactor error handling or something in the future. I think it's better practice to restrict the use of `exit` to the very top-level, e.g. `main` or `run` if `main` is unsuitable for whatever reason.

2. This still returns a success exit code in the case that the rename from `fuelup_new_dir_path.join("fuelup")` to `fuelup_bin` fails. I'm not sure this is what we want?

3. Rather than using `error!` (here and above) and `exit`, perhaps we can format this into an anyhow error that can be returned to keep error handling more consistent?

bingcicle: Hmm, at first the thought was that the `error!`s are there to provide additional info on what `fuelup` is trying to do if the `fs::rename()`s fails. But you're right in that we can just use `bail!` instead with the `error!`s to exit out of the function - updated it to do this instead, with only 1 usage of `error!`

src/ops/fuelup_self.rs
```diff
@@ -1,17 +1,78 @@
...
+
+    info!(
+        "Moving {} to {}",
+        fuelup_new.display(),
+        &fuelup_bin.display()
+    );
+    if let Err(e) = fs::rename(&fuelup_new, &fuelup_bin) {
+        error!(
+            "Failed to replace old fuelup with new fuelup: {}. Attempting to restore backup fuelup.",
+        e);
```
mitchmindtree, line 56: This is such strange formatting, I've never seen this from rustfmt :laughing: That said if rustfmt's happy, I'm happy.

mitchmindtree (approved):
:shipit:

PR #98, opened 2022-07-06
Separate CI step that checks fuelup-init

### LINKED ISSUE #88: fix job that runs `fuelup-init.sh` as a check
There's a small issue with running `fuelup-init.sh` to ensure that fuelup installs properly within the job when releasing `fuelup` - since the script would try to source for the release that was in the process of being released, the binaries were not ready yet. This led to CI failure [here](https://github.com/FuelLabs/fuelup/runs/7018032314?check_suite_focus=true).

This is not a breaking bug however, since it happens only during releases and we can rerun the workflow to have it pass once the binaries are available. We should still fix the behaviour somehow, perhaps by waiting on the binaries to be released if it is ran during a release.

bingcicle:
Fixes #88 

The step that runs `fuelup-init` to check for installation of fuelup should be run at all times, but handled differently:

1) During normal CI checks: it's ok to run it directly within `check-scripts`.
2) During releases: `fuelup-init` has to wait for releases to finish building before attempting to download the `fuelup` binary.

Note that this does not prevent releases from being uploaded if there's an error with installing fuelup-init, but only notifies.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,7 @@ jobs:
 
   try-run-fuelup-init:
     needs: cancel-previous-runs
+    if: github.event_name != 'release' || github.event.action != 'published'
     name: Try fuelup installation with fuelup-init
     strategy:
       matrix:
@@ -47,7 +48,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Attempt to install fuelup through fuelup-init.sh
-        if: github.event_name != 'release' || github.event.action != 'published'
         run: ./fuelup-init.sh
 
   cargo-clippy:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,7 @@ jobs:
           access_token: ${{ github.token }}
 
   check-scripts:
+    needs: cancel-previous-runs
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
@@ -35,6 +36,7 @@ jobs:
           sh_checker_exclude: "src"
 
   try-run-fuelup-init:
+    needs: cancel-previous-runs
     name: Try fuelup installation with fuelup-init
     strategy:
       matrix:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,8 +33,19 @@ jobs:
         with:
           sh_checker_comment: false
           sh_checker_exclude: "src"
+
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -34,6 +34,7 @@ jobs:
           sh_checker_comment: false
           sh_checker_exclude: "src"
       - name: Attempt to install fuelup through fuelup-init.sh
+        if: github.event_name != 'release' && github.event.action != 'published'
```
adlerjohn, line 37: Should this be

```suggestion
        if: github.event_name != 'release' || github.event.action != 'published'
```

adlerjohn, line 37: You should make this step its own job.

.github/workflows/ci.yml
```diff
@@ -250,3 +251,20 @@ jobs:
...
+  post-release-checks:
+    name: Try fuelup install with fuelup-init
+    strategy:
+      matrix:
+        job:
+          - os: ubuntu-latest
+          - os: macos-latest
+    runs-on: ${{ matrix.job.os }}
+    if: github.event_name == 'release' && github.event.action == 'published'
+    needs: [cancel-previous-runs, build-release]
```
adlerjohn, line 264: Move these two number `name` so it's easier to see the conditions this job runs under.

.github/workflows/ci.yml
```diff
@@ -33,7 +34,20 @@ jobs:
...
+    strategy:
+      matrix:
+        job:
+          - os: ubuntu-latest
+          - os: macos-latest
+    runs-on: ${{ matrix.job.os }}
+    steps:
+      - uses: actions/checkout@v3
       - name: Attempt to install fuelup through fuelup-init.sh
+        if: github.event_name != 'release' || github.event.action != 'published'
```
adlerjohn, line 50: There's only a single step here, so maybe move this `if` to be for the whole job?

.github/workflows/ci.yml
```diff
@@ -250,3 +264,20 @@ jobs:
...
+    strategy:
+      matrix:
+        job:
+          - os: ubuntu-latest
+          - os: macos-latest
+    runs-on: ${{ matrix.job.os }}
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v3
+
```
mitchmindtree, line 281:
```suggestion
```

mitchmindtree (dismissed):

mitchmindtree (approved):

PR #103, opened 2022-07-13
curl fuelup-version instead of github api

### LINKED ISSUE #102: curl `fuelup-version` file instead of GitHub API
It seems as though GitHub API might be flaky with fetching the latest version of `fuelup` after some reports of failure with running the `fuelup-init` script with the `fuelup` version somehow missing in the logs.

This issue is also somewhat a followup to #43 so that we can avoid having to curl GitHub API unnecessarily.

bingcicle:
Closes #102
Related #58, #43 

Reduces dependence on GitHub API by querying the published `fuelup-version` instead of querying the latest tag endpoint of GitHub API, since we [publish the version](https://github.com/FuelLabs/fuelup/blob/gh-pages/fuelup-version) already. We still fallback to querying the api endpoint in case curl-ing `fuelup-version` fails.

We also trim the 'v' from the published version within the workflow so that it works nicely with our script.

Note that this change means installing via `fuelup-init` on **master** will break until a new version is published. This should not affect `fuelup-init` that is published onto gh-pages though (which means it should be fine for general public)

### DIFF
--- a/fuelup-init.sh
+++ b/fuelup-init.sh
@@ -24,7 +24,8 @@ main() {
     mkdir -p "$FUELUP_DIR/bin"
 
     local _fuelup_version
-    _fuelup_version="$(curl -s https://api.github.com/repos/FuelLabs/fuelup/releases/latest | grep "tag_name" | cut -d "\"" -f4 | cut -c 2-)"
+    _published_fuelup_version_url="https://raw.githubusercontent.com/FuelLabs/fuelup/gh-pages/fuelup-version"
+    _fuelup_version="$(curl -s $_published_fuelup_version_url)"
     local _fuelup_url="https://github.com/FuelLabs/fuelup/releases/download/v${_fuelup_version}/fuelup-${_fuelup_version}-${_arch}.tar.gz"
 
     local _dir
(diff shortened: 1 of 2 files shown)


bingcicle:
This should be good for review despite the fuelup installation CI check failing, but this is expected because of the updates to the script which will no longer manually `cut` the `v` from the version. The updated script should curl and expect the version without the `v` but the current published version contains a `v`. It should be fine on the next release.

bingcicle:
Just ran into some flakiness in tests for #95 as well that warrants moving to simply curling the hosted file instead of the API for the version. For some reason the script fails to grab the version as well in the CI: [link](https://github.com/FuelLabs/fuelup/actions/runs/2661898575). Had to rerun the job 2 times to get it to work.

fuelup-init.sh
```diff
@@ -24,7 +24,13 @@ main() {
     mkdir -p "$FUELUP_DIR/bin"
 
     local _fuelup_version
-    _fuelup_version="$(curl -s https://api.github.com/repos/FuelLabs/fuelup/releases/latest | grep "tag_name" | cut -d "\"" -f4 | cut -c 2-)"
+    _published_fuelup_version_url="https://raw.githubusercontent.com/FuelLabs/fuelup/gh-pages/fuelup-version"
+    _fuelup_version="$(curl -s $_published_fuelup_version_url)"
+    if echo "$_fuelup_version" | grep -q -E '404|400'; then
```
mitchmindtree, line 29: Would this condition be missed on a future version like `1.0.404`? I guess it's unlikely we'd reach that version :smiling_face_with_tear:  and it would just get fetched from the GitHub API anyway.

mitchmindtree (approved):

PR #107, opened 2022-07-20
add '^' to correctly match compatible files only while indexing versions

bingcicle:
Current channel has published `fuel-core` at `0.9.6` despite failing tests, because of a missing `^` in the regex matching done that doesn't exclude the files marked as `incompatible`.

To confirm:

```console
$ touch incompatible
$ touch compatible
$ ls -1 | grep "compatible*"
compatible
incompatible
$ ls -1 | grep "^compatible*"
compatible
```

Also removed the `*` from the comment.


mitchmindtree (approved):
Good spot!

Issue #69, opened 2022-06-10
`fuelup toolchain install latest` should not fetch / reinstall versions that a user already has fetched / installed


Braqzen:
# Summary

When running `fuelup install` the output states that it first fetches `forc` for the latest version followed by displaying the URL query.
Then it repeats the behaviour for `fuel-core`.

Lastly, it unpacks the plugins.

If I already have a version installed that matches whatever is being fetched it should do one of the following

1. Not fetch it in the first place (it seems to somehow know the latest version _and then_ it fetches it)
2. Not display the installation process for that version / plugin and instead mention that I am up-to-date and there is nothing new that has been installed.

Note: How rustup does it. First time it updates and the second time it syncs and tells me there is nothing to change.

# Screenshots

## fuelup install
![1](https://user-images.githubusercontent.com/103777923/173006408-8f412a2b-44e4-4e2d-9feb-bebba1edf67c.png)

## rustup update
![3](https://user-images.githubusercontent.com/103777923/173007159-246230ce-37ba-407f-b896-9ff597b6cec6.png)

binggh:
This is a problem since fuelup is pretty rudimentary now and merely downloads bins to `.fuelup/bin` 😅 `rustup` stores `update-hashes` within the `.rustup` dir and checks against it to see if the hash exists before deciding if it should download a certain toolchain - will work on this after #73 is completed.

PR #108, opened 2022-07-20
verify download hash

### LINKED ISSUE #110: verify download with sha256 hashes
`fuelup` should compute the hashes of the downloads against the provided hashes within channels to ensure the authenticity of the binaries.

bingcicle:
Closes #110

An upgrade to the existing download function - calculates the hash of the download file and verifies it against the official published channel.


bingcicle:
The channel toml hash should be validated as well - however will be implementing this in #69 since channel hashes will be used to ensure there's no duplicate downloads from the same channel.

adlerjohn (commented):
This one is a bit stale, but is it ready for review? CI is failing.

bingcicle:
> This one is a bit stale, but is it ready for review? CI is failing.

Yup, it should be good for review- was failing because of the changes made in #103 that is now fixed with a hotfix release.

src/download.rs
```diff
@@ -193,12 +200,13 @@ fn unpack(tar_path: &Path, dst: &Path) -> Result<()> {
     Ok(())
 }
 
-pub fn download_file(url: &str, path: &PathBuf) -> Result<()> {
+pub fn download_file(url: &str, path: &PathBuf, hasher: Option<&mut Sha256>) -> Result<()> {
```
mitchmindtree, line 203: It looks like there are no cases where we download a file without hashing it. Maybe we can change this to:

```suggestion
pub fn download_file(url: &str, path: &PathBuf, hasher: &mut Sha256) -> Result<()> {
```

src/download.rs
```diff
@@ -193,12 +200,13 @@ fn unpack(tar_path: &Path, dst: &Path) -> Result<()> {
...
     const RETRY_ATTEMPTS: u8 = 4;
     const RETRY_DELAY_SECS: u64 = 3;
 
     let handle = ureq::builder().user_agent("fuelup").build();
     let mut file = OpenOptions::new().write(true).create(true).open(&path)?;
+    let hasher = RefCell::new(hasher);
```
mitchmindtree, line 209: I don't think there should be any need to use `RefCell`?

src/download.rs
bingcicle: Seems like so for now, so I'll remove it. Initially put it there as an `Option` so that we have the option (pun intended) to use it or not

mitchmindtree (approved):

PR #114, opened 2022-07-23
fuelup check refactor

bingcicle:
Small refactor to make fuelup_check.rs more readable - no logical changes here, just moving things around.

### DIFF
--- a/src/ops/fuelup_check.rs
+++ b/src/ops/fuelup_check.rs
@@ -12,7 +12,7 @@ use crate::{
     fmt::{bold, colored_bold},
     toolchain::{DistToolchainName, Toolchain},
 };
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use semver::Version;
 use termcolor::Color;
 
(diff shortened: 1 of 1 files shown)


src/ops/fuelup_check.rs
```diff
@@ -160,3 +86,86 @@ pub fn check(command: CheckCommand) -> Result<()> {
...
+        match std::process::Command::new(&component_executable)
+            .arg("--version")
+            .output()
+        {
+            Ok(o) => {
+                let version = Version::parse(
+                    String::from_utf8_lossy(&o.stdout)
+                        .split_whitespace()
+                        .nth(1)
+                        .expect("expected version"),
```
mitchmindtree, line 127: I realise we used `expect` previously too, but perhaps it's worth using `ok_or_else` with `anyhow!` instead?

mitchmindtree (approved):
Looks nicer!

I've left one small suggestion, but leaving as optional as it's not particularly related to this PR.

src/ops/fuelup_check.rs
bingcicle: Updated the error handling for this to print an error message instead of aborting.

PR #117, opened 2022-07-25
Use current `latest` channel as `stable` channel instead

### LINKED ISSUE #116: re-define and re-implement `latest` channel
The `latest` channel should consist of the _latest_ binaries of `forc` and `fuel-core`, and the current workflow that we have running tests should produce the `stable` channel instead.

To accomplish this:

1.  Bring back the previous workflow that indexes the latest versions without running tests. This workflow should also run every 30 minutes.
2. Rename the current workflow and tweak configs to create the `stable` channel instead of the `latest` channel

bingcicle:
Closes #116 

My observation of the needs of both internal and external devs seem to suggest that it might be better to always provide the latest binaries regardless of whether they passed tests.

It also makes sense to name the releases that passed integration tests as _compatible_ and therefore _stable_, since they have basically passed all known success and failure tests. Until we have a better definition of what constitutes a _stable_ toolchain, I propose we use this definition in the meantime.

The recent PR that defines the `latest` toolchain #95 will be used as the `stable` toolchain instead while the `latest` toolchain will always provide the `latest` binaries, regardless of whether they passed integration tests.

To accomplish this,

- I brought back parts of the old [index-versions.sh](https://github.com/FuelLabs/fuelup/blob/2c98e0669a0ee4284ac83b701c92dd4694aca88f/.github/workflows/scripts/index-versions.sh) that indexes latest versions indiscriminately as a separate `index-versions-latest.yml` workflow to publish the `latest` channel. This also runs every 30 minutes.
- I made necessary changes to the existing workflow(s) that publishes the `latest` channel, to publish the `stable` channel instead. This is done through changes to names of jobs, files, etc. within the YAML configuration files.
- Updated docs (CHANNELS.md). This should eventually be within an mdbook.

### DIFF
--- a/.github/workflows/index-versions-latest.yml
+++ b/.github/workflows/index-versions-latest.yml
@@ -10,13 +10,14 @@ env:
 
 jobs:
   index-versions:
-    name: Index versions and publish channel (latest)
+    name: Index and publish channel (latest)
     runs-on: ubuntu-latest
     steps:
       - name: checkout master
         uses: actions/checkout@v3
 
-      - name: Get latest versions and update channel if needed
+      - name: Compare latest versions with published channel
+        id: compare-latest-versions
         run: |
           curl -s "https://fuellabs.github.io/fuelup/channel-fuel-latest.toml" -L -o channel-fuel-latest.toml
           mkdir -p ${{ env.LATEST_CHANNEL_DIR }}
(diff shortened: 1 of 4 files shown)


bingcicle:
Closing this PR after a discussion with @mitchmindtree.

`fuelup`'s mission should be to provide tested and reliable toolchains, so therefore we shouldn't need a separately `latest` toolchain that purely has the latest binaries since they might fail integration tests with each other. Any more cutting edge than `latest` should mean either building from source or using custom toolchains (soon ™️)

If we use the mission as guidance for fuelup's development, the changes this PR make would be a regression.

PR #119, opened 2022-07-26
compare-versions: handle 404 error vs other unexpected errors differently

bingcicle:
We should only re-generate the channel toml if we encounter a `404` error. If we encounter some other forms of error, we should retry at the next scheduled run of the CI check.

### DIFF
--- a/scripts/compare-versions/src/main.rs
+++ b/scripts/compare-versions/src/main.rs
@@ -153,7 +153,7 @@ fn main() -> Result<()> {
         }
         Err(e) => {
             bail!(
-                "Unexpected error trying to fetch channel {} - retrying at the next scheduled time",
+                "Unexpected error trying to fetch channel: {} - retrying at the next scheduled time",
                 e
             );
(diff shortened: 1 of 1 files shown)


scripts/compare-versions/src/main.rs
```diff
@@ -150,6 +151,9 @@ fn main() -> Result<()> {
             print_selected_versions(&[latest_sway_version], &[latest_fuel_core_version]);
             std::process::exit(0);
         }
+        Err(_) => {
+            bail!("Unexpected error trying to fetch channel - retrying at the next scheduled time");
```
mitchmindtree, line 155: Should we include the error here so that a user can help us debug? E.g.

```rust
    Err(e) => {
        bail!("Unexpected error trying to fetch channel: {} - retrying at the next scheduled time");
```

mitchmindtree (approved):

adlerjohn (commented):
ci failing

PR #123, opened 2022-07-27
custom toolchains

bingcicle:
custom toolchain support - preliminary work after the discussion within #46. For details, please look at the discussion located in that issue_

Since we only want to distribute __tested__ toolchains as official toolchains, my proposal is to support custom toolchains to allow people to get whichever versions of each component that they need, including the latest ones that failed compatibility tests. Once #104 is merged and released, `fuelup` will always prefer the versions stated in the channel, which means that if new versions are released but fail tests, they will not be made available on the channel. This means we need an intermediate solution to allow people to access latest releases if they want - it would be bad developer experience imo if `fuelup` could not do this, and they had to fallback on `cargo` or even manually downloading the binaries instead. In this case, it seems like our intermediate solution here is something we would want in the future anyway, hence this PR! It's something similar to [`rustup toolchain link`](https://rust-lang.github.io/rustup/concepts/toolchains.html#custom-toolchains) but with a slightly improved usage experience imo.

Not to mention this would help with Sway app developers who have been asking for a way to lock `forc` to a specific version in the CI.

This will be made known as an advanced feature and will mainly be for experimental purposes, documented once I get to #3.

Task list:
- [x] `fuelup toolchain new <name>`: Creates an empty directory named `<name>` with a `/bin` subdirectory, within `.fuelup/toolchains`
- [x] `fuelup default <toolchain>`: Sets default toolchain within `settings.toml`
- [x] `fuelup component add <component>@<version>`: Since the toolchain will be empty at first, the user will have to add components manually eg. `fuelup component add forc@0.18.1`.
- [x] `fuelup component remove <component>`: removes a component

Would love preliminary comments on if this is desirable, or if there's a better way to go about this. cc @mitchmindtree @adlerjohn @FuelLabs/tooling


bingcicle:
I think this should be ready for review - while working on #128 I might run into some bugs/errors that I might retroactively come back here to fix but this should be good to go.

src/ops/fuelup_toolchain/new.rs
```diff
@@ -0,0 +1,40 @@
...
+use std::io;
+
+pub fn new(command: NewCommand) -> Result<()> {
+    let NewCommand { name } = command;
+
+    let toolchain_dir = toolchain_dir();
+
+    let toolchain_exists = fs::read_dir(&toolchain_dir)?
+        .filter_map(io::Result::ok)
+        .filter(|e| e.file_type().map(|f| f.is_dir()).unwrap_or(false))
```
mitchmindtree, line 16: I think this could also be:
```suggestion
        .filter(|e| e.path().is_dir())
```

src/ops/fuelup_toolchain/new.rs
```diff
@@ -0,0 +1,40 @@
...
+        let settings = SettingsFile::new(settings_file);
+        settings.with_mut(|s| {
+            s.default_toolchain = Some(name.clone());
+            Ok(())
+        })?;
+    }
+
+    if ensure_dir_exists(&toolchain_dir.join(toolchain_bin_dir)).is_ok() {
+        println!("New toolchain initialized: {}", &name);
+    };
```
mitchmindtree, line 37: Do we want to swallow the error from `ensure_dir_exists()` here? Maybe we should do:

```rust
ensure_dir_exists(&toolchain_dir.join(toolchain_bin_dir))?;
info!("New toolchain initialized: {}", name);
```

src/ops/fuelup_toolchain/uninstall.rs
```diff
@@ -0,0 +1,21 @@
...
+    let toolchain = Toolchain::from(&name)?;
+
+    if !toolchain.exists() {
+        info!("toolchain '{}' does not exist", &name);
+        return Ok(());
+    }
+
+    if toolchain.uninstall_self().is_ok() {
+        info!("toolchain '{}' uninstalled", &name);
+    };
```
mitchmindtree, line 18: Same here, perhaps we want:
```rust
toolchain.uninstall_self()?;
info!("toolchain '{}' uninstalled", name);
```

src/toolchain.rs
```diff
@@ -1,29 +1,58 @@
...
+use crate::component;
 use crate::download::{download_file_and_unpack, link_to_fuelup, unpack_bins, DownloadCfg};
 use crate::ops::fuelup_self::self_update;
-use crate::path::{ensure_dir_exists, fuelup_bin, fuelup_bin_dir, toolchain_bin_dir};
+use crate::path::{
+    ensure_dir_exists, fuelup_bin, fuelup_bin_dir, settings_file, toolchain_bin_dir,
+};
+use crate::settings::SettingsFile;
+
+pub const RESERVED_TOOLCHAIN_NAMES: &[&str] = &["latest", "nightly"];
```
mitchmindtree, line 16: Perhaps we should conservatively add "stable" and "beta" so that user's toolchains don't break if we eventually introduce them?

Also, it could be useful to have each of these names as `const`s for use independently elsewhere.

src/toolchain.rs
```diff
@@ -1,29 +1,58 @@
...
 
 pub enum DistToolchainName {
     Latest,
 }
 
+impl fmt::Display for DistToolchainName {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            DistToolchainName::Latest => write!(f, "latest"),
```
mitchmindtree, line 25: Like here for example ^

src/toolchain.rs
```diff
@@ -1,29 +1,58 @@
...
+
 impl FromStr for DistToolchainName {
     type Err = anyhow::Error;
     fn from_str(s: &str) -> Result<Self> {
-        let split = s.split_once('-').unwrap();
-        let name = split.0;
+        let name = s.split_once('-').map(|n| n.0);
         match name {
-            "latest" => Ok(Self::Latest),
+            Some("latest") => Ok(Self::Latest),
```
mitchmindtree, line 35: And here ^

src/toolchain.rs
```diff
@@ -1,29 +1,58 @@
...
             _ => bail!("Unknown name for toolchain: {}", s),
         }
     }
 }
 
+pub fn is_official_toolchain(toolchain: &str) -> bool {
+    let mut reserved: Vec<String> = Vec::new();
```
mitchmindtree, line 42: It looks like we shouldn't really need to collect a `reserved` `Vec` here as the `Vec` is never returned and is only used to check if a name exists. Instead, perhaps we can just return directly if we come across a reserved string? E.g. something like:

```rust
for reserved_toolchain_name in RESERVED_TOOLCHAIN_NAMES {
    if toolchain == reserved_toolchain_name {
        return true;
    } else if let Some(triple) = triple {
        if toolchain == reserved_toolchain_name + triple {
            return true;
        }
    }
}
```

Btw just checking - should we be adding a `-` between the toolchain name and the triple?

src/toolchain.rs
```diff
@@ -119,4 +175,35 @@ impl Toolchain {
 
         Ok(download_cfg)
     }
+
+    pub fn remove_component(&self, component: &str) -> Result<()> {
+        if self.has_component(component) {
+            info!("Removing '{}' from toolchain '{}'", component, self.name);
+            let component_path = self.path.join(component);
+            remove_file(component_path)?;
```
mitchmindtree, line 183: Maybe we should provide some context here as I/O errors can be a bit ambiguous on their own, e.g. something like:
```rust
remove_file(component_path)
    .with_context(|| format!("failed to remove component '{}', component))?;
```

src/toolchain.rs
```diff
@@ -119,4 +175,35 @@ impl Toolchain {
...
+            // If component to remove is 'forc', silently remove forc plugins
+            if component == component::FORC {
+                for component in [
+                    component::FORC_FMT,
+                    component::FORC_LSP,
+                    component::FORC_EXPLORE,
+                ] {
+                    let component_path = self.path.join(component);
+                    remove_file(component_path)?;
```
mitchmindtree, line 192: Same here

mitchmindtree: Also - are we sure these plugins will always exist? E.g. in the near future we may add `FORC_WALLET` or `FORC_CLIENT` to this list. Perhaps we should first check `component_path.exists()`?

src/toolchain.rs
```diff
@@ -119,4 +175,35 @@ impl Toolchain {
...
+            info!("Removing '{}' from toolchain '{}'", component, self.name);
+            let component_path = self.path.join(component);
+            remove_file(component_path)?;
+            // If component to remove is 'forc', silently remove forc plugins
+            if component == component::FORC {
+                for component in [
+                    component::FORC_FMT,
+                    component::FORC_LSP,
+                    component::FORC_EXPLORE,
+                ] {
```
mitchmindtree, line 190: Maybe we should consider moving this list to a `const` array, e.g. `component::FORC_PLUGINS`. It will be important to have an easy-to-find, central source of truth for this list as we add new plugins.

mitchmindtree (changes requested):
Overall looking good! Just a few small things to address.

I can imagine it might also be useful to add a `list` command for both `toolchain` and `component` for inspecting what toolchains are installed, what components are under each and at what version etc. We can probably leave this for a future PR though as this one's already big enough and it's not essential.

Cargo.toml
```diff
@@ -19,7 +19,7 @@ path = "src/main.rs"
 
 [dependencies]
 anyhow = "1"
-clap = { version = "3.1", features = ["cargo", "derive", "env"] }
+clap = { version = "3.2", features = ["cargo", "derive", "env"] }
```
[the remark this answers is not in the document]
bingcicle: Yup - we're using `value_parser` to validate/parse custom toolchain names. and that clap feature received major changes in `3.2` over `3.1` [release notes](https://github.com/clap-rs/clap/releases/tag/v3.2.0)

src/commands/component.rs
```diff
@@ -0,0 +1,33 @@
+use anyhow::Result;
+use clap::Parser;
+
+use crate::ops::fuelup_component::{add::add, remove::remove};
+
+#[derive(Debug, Parser)]
+pub enum ComponentCommand {
+    /// Add a component to component to the currently active Fuel toolchain
```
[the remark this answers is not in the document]
bingcicle: Good ideas - and thanks for the typo spot

src/toolchain.rs
bingcicle: Ah yes - we should be, good spot here

src/commands/default.rs
```diff
@@ -0,0 +1,40 @@
...
+    let DefaultCommand { toolchain } = command;
+
+    let current_toolchain = Toolchain::from_settings()?;
+
+    if toolchain.is_none() {
+        println!("{} (default)", current_toolchain.name);
+        return Ok(());
+    };
+
+    let toolchain = toolchain.unwrap();
```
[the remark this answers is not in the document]
bingcicle: Ah yes, wrote most of these before we had that conversation about whether to keep using `tracing` and didn't switch the `println!`s back to `info!`s

src/commands/component.rs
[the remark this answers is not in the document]
bingcicle: >Maybe we should also add a note stating that we encourage using the standard, integration-tested toolchains if possible and to only use a custom toolchain if the user has a special use-case?

Should this be included within the fuelup book instead?

src/commands/default.rs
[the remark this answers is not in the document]
bingcicle: Addressed most of the `println!`s but there are still some remaining that are unrelated to this PR - will open a separate PR for those

src/toolchain.rs
bingcicle: Actually, it seems like I stopped using this for whatever reason. I actually do not recall but it might've been for the `fuelup toolchain new <toolchain>` command. I'm going to remove this for now

src/toolchain.rs
bingcicle: Just remembered there was a `SUPPORTED_PLUGINS` that i created a while back that I can use

src/toolchain.rs
bingcicle: in the entrypoint of `remove_component` we first check that `self.has_component`, which should handle checking that the path exists. So this shouldn't be a worry

mitchmindtree (approved):
Nice one, looking good!

It looks like CI is failing, though I don't recall if this error is expected?

> fuelup: fuelup v0.2.2 was not found - either the release is not ready yet or the tag is invalid. You can check if the release is available here: https://github.com/FuelLabs/fuelup/releases/vv0.2.2

At the very least, it looks like we have an error with the error message construction - notice the `vv0.2.2` should probably be `v0.2.2`. Possibly related to the failure?

bingcicle:
> Nice one, looking good!
> 
> It looks like CI is failing, though I don't recall if this error is expected?
> 
> > fuelup: fuelup v0.2.2 was not found - either the release is not ready yet or the tag is invalid. You can check if the release is available here: https://github.com/FuelLabs/fuelup/releases/vv0.2.2
> 
> At the very least, it looks like we have an error with the error message construction - notice the `vv0.2.2` should probably be `v0.2.2`. Possibly related to the failure?

It's the same error with all the last few PRs (and currently open PRs) because I made this change: https://github.com/FuelLabs/fuelup/pull/103

I think I will cut a new release up until and excluding the channel-related changes just so we can avoid this CI failure.

adlerjohn (approved):
utACK

PR #126, opened 2022-07-28
`Settings` serialization/deserialization

### LINKED ISSUE #115: refactor `Settings` to derive `Deserialize`
As mentioned [here](https://github.com/FuelLabs/fuelup/pull/104#pullrequestreview-1048861769) in #104 

Refactor `Settings` type to derive `Deserialize` rather than doing manual parsing.

bingcicle:
closes #115 

- Implements serialization/de-serialization for `Settings`
- Added tests for reading/writing `Settings`
- [x] parse `Settings`
- [x] write into settings file
- [x] stringify `Settings`

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -1,7 +1,8 @@
+use serde::{Deserialize, Serialize};
 use std::{cell::RefCell, path::PathBuf};
-use toml_edit::{value, Document, Table, Value};
+use toml_edit::{de, ser, value, Document, Table};
 
-use anyhow::{bail, Result};
+use anyhow::Result;
 
 use crate::file;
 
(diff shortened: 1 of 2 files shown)


src/settings.rs
```diff
@@ -61,56 +62,85 @@ impl SettingsFile {
...
-fn get_opt_string(table: &mut Table, key: &str) -> Result<Option<String>> {
-    if let Ok(v) = take_value(table, key) {
-        if let Value::String(s) = v {
-            Ok(Some(s.to_string()))
-        } else {
-            bail!("Expected string, got {}", key)
-        }
-    } else {
-        Ok(None)
+    pub(crate) fn stringify(self) -> Result<String> {
```
mitchmindtree, line 76: `to_string` would be a little more conventional, or even `to_toml_string`

mitchmindtree (approved):
Nice, much cleaner! One minor nit but nothing major, will leave it up to you :shipit:

bingcicle:
hang on, clippy nit

src/settings.rs
bingcicle: I changed it to `to_string` here. Also noticed (thanks to clippy) since `ser::to_document` takes a reference to self, we can therefore change `into_toml` to `to_toml` instead

mitchmindtree (approved):
Looking good! I've disabled auto-merge as it looks like something is up with CI?

bingcicle:
> Looking good! I've disabled auto-merge as it looks like something is up with CI?

The fuelup installation check is broken until we release a new version because of #103

PR #128, opened 2022-08-04
integration tests

### LINKED ISSUE #129: Integration tests
with #123 now in the works, `fuelup` has a number of commands to be used now. We should get integration tests up and running that basically attempt to replicate actual usage of `fuelup` and check that every command is working as intended.

bingcicle:
Closes #129 

Sets up a sandboxed `.fuelup` directory with the necessary directories/files using `tempfile` crate to ensure auto cleanup after the tests are done (eg. `settings.toml`, `/toolchains`), along with a sandboxed `HOME` environment variable for a copy of the `fuelup` binary to act on. This _should_ be thread safe since we are saving paths within a `TestConfig`, using each one for one test and each command will only run the specific binary which acts on their own created sandbox.

There is also the concept of a `FuelupState` - this will pre-create any directories within the sandboxed `.fuelup` directory before we run commands on it. These will just be file stubs for now that are created in the [tempdir](https://docs.rs/tempfile/3.3.0/tempfile/fn.tempdir_in.html) and is automatically cleaned up after, and in future if there are more scenarios to setup we will expand `FuelupState`.

For now we _have_ to run tests on a network so that we can ensure `fuelup toolchain install` works as intended, unless we come up with some way to mock installation (which we probably eventually want to do)

Command tests:
- [x] --version
- [x] toolchain install
- [x] check
- [x] self update
- [x] default (empty)
- [x] toolchain new
- [x] toolchain new disallow official toolchain names
- [x] toolchain new and set default
- [x] component add
- [x] component add specific version disallowed for official toolchains
- [x] component remove


bingcicle:
This should also be good to go alongside #123

tests/testcfg.rs
```diff
@@ -0,0 +1,107 @@
...
+            fs::File::create(&bin_dir.join("forc-explore"))?;
+            fs::File::create(&bin_dir.join("fuel-core"))?;
+            fs::copy(
+                &env::current_dir()
+                    .unwrap()
+                    .join("tests/settings-example.toml"),
+                &tmp_fuelup_root_path.join("settings.toml"),
+            )
+            .expect("Failed to copy settings");
+        }
```
mitchmindtree, line 97: Is the idea behind this `LatestToolchainInstalled` branch to emulate having `fuelup` in a state where `latest` is already installed? Could you elaborate a bit on why we need this, and why we manually create stub dirs rather than doing `fuelup toolchain install latest` to have it auto-populate with the binaries etc too?

mitchmindtree (commented):
Looking good! Lots of nice sanity checks in there, and that `TestCfg` abstraction is looking very handy too, it makes the tests look super clean and focused :ok_hand:

tests/testcfg.rs
bingcicle: Yes, the main idea is to emulate `fuelup` in a state where `latest` is already installed. I think it'd be ideal to have as little tests that rely on the network as possible, and hence that is also why I combined the `install` and `check` tests into the same test.

In the case of certain commands like `fuelup component remove <component>`, downloading seems like an unnecessary step, since all we want to do is to remove something called `forc` for example and then ensure that it isn't there, without deleting other components. In this case we can just have a file stub called `forc` and we wouldn't need to download the actual `forc`. Same for when uninstalling toolchains - we just want to make sure that a toolchain directory is no longer there.

bingcicle: An alternative is to download the binaries once at the start of `cargo test` and reuse them whereever needed by copying them into the required toolchain

mitchmindtree: > In the case of certain commands like fuelup component remove <component>, downloading seems like an unnecessary step, since all we want to do is to remove something called forc for example and then ensure that it isn't there, without deleting other components. In this case we can just have a file stub called forc and we wouldn't need to download the actual forc. Same for when uninstalling toolchains - we just want to make sure that a toolchain directory is no longer there.

Yeah I can empathise with wanting to reduce network I/O for tests. My concern is that we end up testing for a case that can never occur anyway, and it might end up being more of a pain to synchronise this set of stubs with the actual behaviour of fuelup as we add different plugins and so-on :thinking: 

> An alternative is to download the binaries once at the start of cargo test and reuse them whereever needed by copying them into the required toolchain

Ahh this would be nice, though I'm not sure how easy it would be to do this in a manner that still allows for separate tests? I guess we could have a single `#[test] fn integration() { ... }` or something like that which downloaded first and then ran the other tests as functions in order one at a time? It would be a shame to not have an individual `#[test]` for each command, but perhaps worth it to remove the need for all the extra networking and to avoid the need to fake a fuelup installation with stubs?

bingcicle: >An alternative is to download the binaries once at the start of cargo test and reuse them whereever needed by copying them into the required toolchain

Perhaps this isn't possible after all and I spoke too soon? Not sure if setting everything up before all tests would be possible. Seems like it is still an open RFC (https://github.com/rust-lang/rfcs/issues/1664).

> I guess we could have a single #[test] fn integration() { ... } or something like that which downloaded first and then ran the other tests as functions in order one at a time? It would be a shame to not have an individual #[test] for each command, but perhaps worth it to remove the need for all the extra networking and to avoid the need to fake a fuelup installation with stubs?

Hmmm.. I guess this would work as a workaround but it just seems like a tradeoff between the code complexity vs. faking installation with file stubs... in which case it feels like stubs are a 'good-enough' solution IMO since in most cases (for now) we would only want to test the file/directory based manipulation around managing toolchains/components... unless we eventually want to test advanced features. For example, features that would require us to actually execute the real binaries and check the stdout/stderr of what `forc` produces, for example. But it seems to me that testing fuelup's role in such an example would just be testing if the correct executable is called.

mitchmindtree: OK let's see how these tests go for now. If it turns out these stubs end up causing too much churn we can revisit this in the future.

mitchmindtree (commented):
Looks like CI failure related to the placeholder toolchain name in the stub?

tests/testcfg.rs
bingcicle: I think it's a result of the error not piping to stderr for some weird reason though it works on my local tests. My guess that it somehow was printed to ardour instead in the CI. I haven't figured out why though...

bingcicle:
> Looks like CI failure related to the placeholder toolchain name in the stub?

I think it's the result of the errors piping out to stdout instead of stderr for some reason, because the tests pass locally. I haven't figured out why though.

JoshuaBatty (approved):
Nice work @bingcicle, looking good.

PR #131, opened 2022-08-10
Bump to v0.3.0

bingcicle:
This new version should contain features to support custom toolchains, as well as various QoL changes to the existing fuelup API. Since supporting custom toolchains resulted in almost twice as many commands as before being created - we will want to have integration testing as well to ensure stability of the tool, before I am comfortable with cutting this release.
We should also be documenting the changes in an mdbook.

Depends on:
- [x] #123
- [x] #128 
- [x] #134 
- [x] #140


bingcicle:
Fixing the tests before releasing this.

kayagokalp (approved):

bingcicle:
thanks @kayagokalp ! Will merge this and cut the v0.3.0 release in a few hours when I'm awake so I can fix any issues that crop up

PR #137, opened 2022-08-16
Refactor usage of `TargetTriple`

### LINKED ISSUE #127: Remove usage of `target_from_name`
Initially `TargetTriple` wasn't scoped out yet but now that we're reusing `target_from_name` everywhere, we should use the `TargetTriple` struct instead of `Option<String>` as input to `DownloadCfg`.

bingcicle:
Closes #127 

- `target_from_name` was a very vague name. Refactored it to `from_component` and place it under `impl TargetTriple`, since the main purpose of that function was to format the `TargetTriple` string from the appropriate component given (eg. forc binaries = `linux_amd64` versus `fuel-core` = `x86_64-linux-gnu`)

Note that the logic within the refactored fn did not change; only the naming and the usage of it.

Also,

- Made `DownloadCfg` struct accept `TargetTriple` for `target` instead of `String`
- Moved all the code related to `TargetTriple` into its own dedicated file `target_triple.rs`


src/toolchain.rs
```diff
@@ -50,45 +51,10 @@ pub struct Toolchain {
...
-        };
-
-        let target_triple = format!("{}-{}-{}", architecture, vendor, os);
-
-        Ok(Self(target_triple))
-    }
-}
-
 impl Toolchain {
     pub fn new(name: &str, target: Option<String>) -> Result<Self> {
```
mitchmindtree, line 55:
```suggestion
    pub fn new(name: &str, target: Option<TargetTriple>) -> Result<Self> {
```
Maybe we should take a `TargetTriple` as an argument here now that we have the type for it?

mitchmindtree (dismissed):
Nice! I've left a couple of minor optional suggestions but neither are worthy of blocking this.

src/toolchain.rs
bingcicle: I decided to remove the `target` arg instead after re-looking at its usage, because right now we are only passing `None` anyway. I was thinking to use an `Option` since in future we might want to support installation of cross-platform toolchains, but we can always extend this in future when that use-case comes around.

src/target_triple.rs
```diff
@@ -0,0 +1,97 @@
...
+pub struct TargetTriple(String);
+
+impl fmt::Display for TargetTriple {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+impl TargetTriple {
+    pub fn new(name: &str) -> Self {
```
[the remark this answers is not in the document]
bingcicle: Removed this altogether as well, since the only place of usage was in `name_allowed` for comparison (which we can use `to_string()` to compare anyway)

mitchmindtree (approved):

PR #138, opened 2022-08-17
publish workflow run URL when `test-toolchain-compatibility` runs

### LINKED ISSUE #124: Implement a way to link channel deployments to specific workflow runs
Currently, workflow runs that test version compatibility may be buried within a lot of similarly named `Check Versions`. It may get difficult to find the specific [workflow](https://github.com/FuelLabs/fuelup/runs/7552214908?check_suite_focus=true) that actually ran the tests, and the [gh-pages commit](https://github.com/FuelLabs/fuelup/commit/c8f8d3bc5206f5774ecef6ecf069bd2b81472d02) do not provide info on this. It would be nice if there was some way to link to the specific workflow run within the commit message.

bingcicle:
Closes #124 

Previously we had no way of knowing the source workflow run when releases were published and `test-toolchain-compatibility.yml` workflow ran. Action runs stack up because they are checked every 30 minutes, so unless we had a direct link to the workflow run or at least the date of the release, it is difficult to trace test failures back to the specific run to investigate further.

In this PR we make use of `GITHUB_RUN_ID` in the [GitHub default environment vars](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables) to include the link to the specific run when we publish an incompatibility or re-publish the channel.

You can see an example of how this might look like within the [sample channel TOML](https://github.com/FuelLabs/fuelup/blob/013c8c532d7ffe6eaa354e4eb97d7b62f42a518a/tests/channel-fuel-latest-example.toml) used in the tests


mitchmindtree (approved):
Great idea!

PR #140, opened 2022-08-18
fix integration tests

bingcicle:
Some inconsistencies in the tests due to bumping the version of `fuelup` that wasn't considered before.

1) we assert `success()` status in the `fuelup_check` now instead of the exact version string.
2) remove the `fuelup` version string check from the toolchain installation test.


bingcicle:
`fuel-core` is not ready to be uncommented yet since within the `latest` channel `0.9.8` is downloaded (which has the version error bug)

JoshuaBatty (approved):

kayagokalp (approved):

Braqzen (approved):

PR #146, opened 2022-08-19
Update supported components

### LINKED ISSUE #142: Add the forc-client plugin commands, `forc deploy`, `forc run`

This is the final step for https://github.com/FuelLabs/sway/issues/2412.

### Solution you'd like

Ensure `forc-client` plugin is shipped by default along with the rest of the forc plugins.

### Notes

Some consideration might be needed to distinguish between the "plugin" (`forc-client`) and the two command binaries it provides (`forc deploy`, `forc run`).

We should prioritise this to make sure these commands are still available on the next published release (or soon after at least).

bingcicle:
Closes #142


kayagokalp (approved):
Lgtm

PR #151, opened 2022-08-20
better handling of 'fuelup check' downgrade behaviour

### LINKED ISSUE #150: `fuelup check` tells you to downgrade `fuel-core`
### Problem

As described in the title. This happens only when you already have a version of a binary later than what comes published within the channel.

### Steps

1. Have a component installed prior to `fuelup v0.3` since it didn't use channels before. In this case, `fuel-core v0.10.1` works.
2. Run `fuelup self update` to get `fuelup v0.3`
3. Run `fuelup check`, this erroneously shows what is published in the channel (in this case, `fuel-core v0.9.8`)

### Possible Solution(s)

_No response_

### Notes

_No response_

### Fuelup version

```console
0.3.0
```

### Installed components

```console
...
fuel-core - Update available: 0.10.1 -> 0.9.8
...
```

bingcicle:
Fixes #150 

Previously `fuelup check` assumes that you would never be able to download a version greater than what's published on a channel. In the upgrade from `fuelup v0.2` to `fuelup v0.3`, there were some users that upgraded `fuel-core` to `v0.10.1`, while the channel still has `v0.9.8` published (due to failing compatibility tests). This resulted in `fuelup check` showing that there's an update available from the more recent version to the older version.

This PR handles the above issue by:

1) Explicitly telling the user that the newer version is 'unstable', and the older version is 'recommended', though this is an edge case that won't happen unless the user manually moved a newer binary into the toolchain (since `fuelup check` only checks official toolchains).

2) There's also some QoL changes by extracting the common code that prints the version information and using `Ordering` for a cleaner look instead.

Example:

Ordering

### DIFF
--- a/src/ops/fuelup_check.rs
+++ b/src/ops/fuelup_check.rs
@@ -92,7 +92,7 @@ fn check_fuelup() -> Result<()> {
     ) {
         bold(|s| write!(s, "{} - ", component::FUELUP));
         compare_versions(
-            &Version::parse(&FUELUP_VERSION)?,
+            &Version::parse(FUELUP_VERSION)?,
             &fuelup_download_cfg.version,
         );
     } else {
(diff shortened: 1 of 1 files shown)


mitchmindtree (approved):
:+1:

PR #154, opened 2022-08-22
forc client check

### LINKED ISSUE #145: Update `fuelup check --verbose` to print `forc-client`
~`fuelup check --verbose` currently does not show `forc-client` version.~

Correction: it does show the versions of `forc-deploy` and `forc-run` separately:

Screenshot 2022-08-19 at 8 18 14 PM
 
Do we want to keep this, or combine both using logic in `fuelup check` into `forc-client`?

bingcicle:
Closes #145 

Small update to `fuelup check --verbose` to correctly display `forc-run` and `forc-deploy` as the same package `forc-client`.

### DIFF
--- a/src/ops/fuelup_check.rs
+++ b/src/ops/fuelup_check.rs
@@ -165,6 +165,12 @@ fn check_toolchain(toolchain: &str, verbose: bool) -> Result<()> {
 
         if verbose && component == component::FORC {
             for plugin in SUPPORTED_PLUGINS {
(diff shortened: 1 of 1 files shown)


mitchmindtree (approved):
Looks OK as a temporary solution, though we may want a more general solution in the future as we add more plugins that support multiple command binaries, e.g. https://github.com/FuelLabs/sway/issues/2369.

bingcicle:
Sounds good - I'll create an issue for this just so we remember!

PR #157, opened 2022-08-22
Fix `toolchain install` command

bingcicle:
https://github.com/FuelLabs/fuelup/commit/54772294712f658255de44751e74e23aa51acc56

This was an unintended change that caused `toolchain install` to look for `forc-lsp` - not sure how I didn't catch this but seems like this was accidentally updated. 

Should be immediately followed up with a new patch release since this is an urgent bug fix.


mitchmindtree (approved):
Good spot!

PR #167, opened 2022-08-30
nightly support

### LINKED ISSUE #159: update `fuelup` to support downloading from `nightly` channel
part of #80 
prereq #158

bingcicle:
Closes #159

Allows for the command `fuelup toolchain install nightly` which sources the channel TOML from gh-pages ([example](https://github.com/FuelLabs/fuelup/blob/b0daef81746066a7c507e784eedad3f0bfc18484/channel-fuel-nightly.toml)) and downloads from the links there.

Note that this does not yet allow for specifying dates or targets, even though some of the infra code that allows for it is in this PR.

In order to support nightly, there were some refactoring that had to be done:
-  `Package` struct now takes version as a `String` instead of a `Version`. This is because nightly versions are semver versions postfixed with `-nightly` and the date at the back. `fuelup check` will use this date to check for updates.
- We used to download the channel TOMLs as temporary files, but this PR will install them into the toolchain directory i.e. `toolchains/<my-toolchain>/ `. Previously we executed the actual binaries to check versions but with nightlies the semver versions are the same if we check nightlies of the same `latest` release. We use the channels to compare versions instead.
- As mentioned, much of the infra code to support parsing of date and target when we install toolchains is in this PR to support specifying a specific nightly to install, but is left unused for now.
- Refactored some paths.

Example `fuelup check` 

Screenshot 2022-09-01 at 12 08 24 AM

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -1,18 +1,15 @@
 use crate::{
-    constants::{
-        CHANNEL_LATEST_FILE_NAME, CHANNEL_NIGHTLY_FILE_NAME, DATE_FORMAT, FUELUP_GH_PAGES,
-    },
+    constants::{CHANNEL_LATEST_FILE_NAME, CHANNEL_NIGHTLY_FILE_NAME, FUELUP_GH_PAGES},
     download::{download_file, DownloadCfg},
     file::read_file,
     toolchain::{DistToolchainName, OfficialToolchainDescription},
 };
 use anyhow::{bail, Result};
 use semver::Version;
-use serde::{de::Visitor, Deserialize};
+use serde::{Deserialize, Deserializer};
 use sha2::{Digest, Sha256};
-use std::fmt;
 use std::{collections::HashMap, path::PathBuf};
-use time::Date;
+use time::{macros::format_description, Date};
 use toml_edit::de;
 
 pub const LATEST: &str = "latest";
@@ -34,49 +31,36 @@ pub struct Channel {
 #[derive(Debug, Deserialize)]
 pub struct Package {
     pub target: HashMap<String, HashedBinary>,
-    pub version: Version,
-    pub date: Option<Date_>,
+    #[serde(deserialize_with = "version_and_date_from_str")]
+    pub version: PackageVersion,
 }
 
-#[derive(Debug, Eq, PartialEq)]
-pub struct Date_(Date);
-
-impl fmt::Display for Date_ {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        self.0.fmt(f)
-    }
+#[derive(Debug)]
+pub struct PackageVersion {
+    pub core: Version,
+    pub date: Option<Date>,
 }
 
-struct DateVisitor;
+fn version_and_date_from_str<'de, D>(deserializer: D) -> Result<PackageVersion, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let s = String::deserialize(deserializer)?;
+    let (version, date) = match s.split_once(' ') {
+        Some((version, date)) => (version, Some(date)),
+        None => (s.as_str(), None),
+    };
 
-impl<'de> Visitor<'de> for DateVisitor {
-    type Value = Date_;
+    let parsed_date = if let Some(date) = date {
+        Date::parse(date, format_description!("([year]-[month]-[day])")).ok()
+    } else {
+        None
+    };
 
-    fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
-        formatter.write_str("a date in the format (YYYY-MM-DD)")
-    }
-
-    fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
-    where
-        E: serde::de::Error,
-    {
-        Ok(Date_(
-            Date::parse(
-                v.split_once('(').unwrap().1.trim_end_matches(')'),
-                DATE_FORMAT,
-            )
-            .unwrap(),
-        ))
-    }
-}
-
-impl<'de> Deserialize<'de> for Date_ {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-    where
-        D: serde::Deserializer<'de>,
-    {
-        deserializer.deserialize_str(DateVisitor)
-    }
+    Ok(PackageVersion {
+        core: Version::parse(version).unwrap(),
+        date: parsed_date,
+    })
 }
 
 impl Channel {
@@ -141,12 +125,12 @@ mod tests {
         assert_eq!(channel.pkg.keys().len(), 2);
         assert!(channel.pkg.contains_key("forc"));
         assert_eq!(
-            channel.pkg["forc"].version,
+            channel.pkg["forc"].version.core,
             Version::parse("0.17.0").unwrap()
         );
         assert!(channel.pkg.contains_key("fuel-core"));
         assert_eq!(
-            channel.pkg["fuel-core"].version,
+            channel.pkg["fuel-core"].version.core,
             Version::parse("0.9.4").unwrap()
         );
 
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -37,8 +37,8 @@ pub struct Package {
     pub date: Option<Date_>,
 }
 
-#[derive(Debug)]
-struct Date_(Date);
+#[derive(Debug, Eq, PartialEq)]
+pub struct Date_(Date);
 
 struct DateVisitor;
 
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -5,10 +5,10 @@ use crate::{
     toolchain::{DistToolchainName, OfficialToolchainDescription},
 };
 use anyhow::{anyhow, bail, Context, Result};
-use core::fmt;
 use semver::Version;
 use serde::{Deserialize, Deserializer};
 use sha2::{Digest, Sha256};
+use std::fmt;
 use std::str::FromStr;
 use std::{collections::HashMap, path::PathBuf};
 use time::{macros::format_description, Date};
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -1,12 +1,14 @@
 use crate::{
-    constants::{CHANNEL_LATEST_FILE_NAME, CHANNEL_NIGHTLY_FILE_NAME, FUELUP_GH_PAGES},
+    constants::{
+        CHANNEL_LATEST_FILE_NAME, CHANNEL_NIGHTLY_FILE_NAME, DATE_FORMAT, FUELUP_GH_PAGES,
+    },
     download::{download_file, DownloadCfg},
     file::read_file,
     toolchain::{DistToolchainName, OfficialToolchainDescription},
 };
 use anyhow::{bail, Result};
 use semver::Version;
-use serde::Deserialize;
+use serde::{de::Visitor, Deserialize};
 use sha2::{Digest, Sha256};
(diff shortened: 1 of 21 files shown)


bingcicle:
Not ready just yet - missing a little something from `fuelup show` not showing the dates within the version.

bingcicle:
So, there's a behaviour in `component add` that I'm not quite sure how to resolve - prior to `nightly` support we're adding the `latest` version of components. It still does that regardless of if you're on `nightly`. I'd like to update this behaviour so if `fuelup` detects that you were using `nightly` version of a component, doing `fuelup component add forc` adds the __latest nightly__ version of `forc`. Otherwise we take the __latest__ from the __latest__ channel as the default to fetch versions from.

However I can see how this might result in confusion (especially when using official toolchains, considering now that we have more than 1). I'm thinking if it might be a better idea to disallow manually adding/removing components completely in official toolchains (there wouldn't be a reason to do that anyway, unless someone can think of a reason?)

This way we can keep the default behaviour of always adding `latest` versions of bins when adding a component without specifying a version on __custom__ toolchains. Then in future, we can support syntax like:


`fuelup component add forc@nightly`
`fuelup component add forc@nightly-2022-08-31`.

bingcicle:
Have disallowed it in the latest commit - `fuelup component add` and `fuelup component remove` disallowed for __official__ toolchains.

bingcicle:
Since the nightly just published I'm able to show what it looks like when I do `fuelup check`:

Screenshot 2022-09-01 at 9 49 43 AM

mitchmindtree (commented):
Mostly looking good - feel free to re-request a review once you're finished with the amendments you mentioned!

mitchmindtree:
Ahh just noticed your new comments about adding/removing components.

> However I can see how this might result in confusion (especially when using official toolchains, considering now that we have more than 1). I'm thinking if it might be a better idea to disallow manually adding/removing components completely in official toolchains (there wouldn't be a reason to do that anyway, unless someone can think of a reason?)

Do you recall what `rustup` does here?

I suspect that adding / removing components will always use components of the current toolchain's channel. I.e. if you're on `latest`, always add/remove `latest` components. If you're on `nightly`, always add/remove nightly components. For specific versions of components, I imagine we only support this on custom toolchains, which I believe is already the case?

src/download.rs
```diff
@@ -34,7 +35,7 @@ struct LatestReleaseApiResponse {
 pub struct DownloadCfg {
     pub name: String,
     pub target: TargetTriple,
-    pub version: Version,
+    pub version: String,
```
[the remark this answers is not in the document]
bingcicle: Hmm, on a second look I think your suggestion is be better. I missed out the fact that `semver` crate allows parsing of the prerelease identifier as well so `0.22.1-nightly` is a valid `Version`. 

However I'd like to tweak your suggestion a bit and do something like this:

```rust
struct DownloadCfg {
  pub version: Version, // parse either `0.21.2` or `0.21.2-nightly`
  pub date: Option<Date> // have an optional date field for nightlies
}

// Same for the Package struct which deserializes from a Channel
struct Package {
  pub target: HashMap<String, HashedBinary>,
  pub version: Version,
  pub date: Option<Date>
}
```

Would this be an OK solution you think? This way we don't have to maintain 2 different `Version`s.

bingcicle:
> Do you recall what `rustup` does here?

Yeah, I tested it out and it seems like they download the latest version of the components of whichever channel they're on. But I'm suspecting this feature is because of the non-required components of their toolchain like `miri` which isn't packaged together by default(?)

> I imagine we only support this on custom toolchains, which I believe is already the case?

No, I made it so that adding `forc` for example automatically added the `forc` within the `latest` channel. This wasn't an issue before since we only had 1 channel, but this is something we might have to consider now, and looking back it seems like it isn't good to allow users to edit the pre-packaged official toolchains anyway (unless we have optional components in future like Rust's `miri`).

src/download.rs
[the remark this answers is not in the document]
bingcicle: So I wrote the above without consideration the magic behind `serde`. Turns out I can do something like:

```rust
struct Version {
    core: Version,
    date: Option<Date>,
}
```
(note: core is something I got from the [semver spec](https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions))

and write a custom deserialization function to create an optional Date. This makes it easier for `fuelup check` as well because we can simply decide whether to compare dates solely based on if it's `Some` (without needing to check the toolchain it's running on)

src/download.rs
```diff
@@ -70,21 +78,34 @@ impl DownloadCfg {
...
             hash,
         })
     }
 }
 
-pub fn tarball_name(name: &str, version: &Version, target: &TargetTriple) -> Result<String> {
+pub fn tarball_name(
+    name: &str,
+    version: &Version,
+    date: Option<Date>,
```
mitchmindtree, line 100: Should this accept a `&PackageVersion` instead of version and date separately?

src/download.rs
```diff
@@ -70,21 +78,34 @@ impl DownloadCfg {
...
+    target: &TargetTriple,
+) -> Result<String> {
+    let version = if let Some(d) = date {
+        version.to_string() + "-" + &d.to_string()
+    } else {
+        version.to_string()
+    };
```
mitchmindtree, line 107: Could we potentially be using a `Display` impl for `PackageVersion` here instead? E.g.
```rust
let version_str = format!("{}", version);
```

src/download.rs
```diff
@@ -35,13 +36,19 @@ pub struct DownloadCfg {
     pub name: String,
     pub target: TargetTriple,
     pub version: Version,
+    pub date: Option<Date>,
```
mitchmindtree, line 39: Maybe we should use a `PackageVersion` rather than separate `version` and `date` fields here?

src/ops/fuelup_check.rs
```diff
@@ -3,164 +3,168 @@ use crate::{
...
+use tempfile::tempdir_in;
 use termcolor::Color;
+use time::Date;
 use tracing::error;
 
 use crate::{component, download::DownloadCfg};
 
-fn collect_versions(channel: Channel) -> HashMap<String, Version> {
-    let mut latest_versions: HashMap<String, Version> = HashMap::new();
+fn collect_versions_and_dates(channel: Channel) -> HashMap<String, (Version, Option<Date>)> {
```
mitchmindtree, line 29:
```suggestion
fn collect_pkg_versions(channel: Channel) -> HashMap<String, PackageVersion> {
```
^ could this make sense here?

src/ops/fuelup_check.rs
```diff
@@ -3,164 +3,168 @@ use crate::{
...
     latest_versions
 }
 
-fn compare_and_print_versions(current_version: &Version, target_version: &Version) {
-    match current_version.cmp(target_version) {
+fn compare_and_print_versions(
+    current_version: &Version,
+    current_date: Option<Date>,
+    latest_version: &Version,
+    latest_date: Option<Date>,
```
mitchmindtree, line 43: Another use case for two `PackageVersion` args here?

mitchmindtree (commented):
I've made a start on reviewing here, though it looks like there are lots of places we are passing separate `version` and `date` fields/arguments where we could instead be using your new `PackageVersion` type? Let me know your thoughts!

bingcicle:
>it looks like there are lots of places we are passing separate `version` and `date` fields/arguments where we could instead be using your new `PackageVersion` type?

Yeah, that didn't occur to me but you're right. Going to refactor the usages everywhere.

src/download.rs
bingcicle: Hmm, the tarball names for nightlies actually don't include the semver [(link)](https://github.com/FuelLabs/sway-nightly-binaries/releases), eg. `forc-binaries-nightly-<date>-<target>.tar.gz` (where `date` is in the format `YYYY-MM-DD`) . So a `Display` impl wouldn't work here since we would just want `nightly-<date>`, and we'd have to handle this case differently.

Do we want to include the semver in the releases for nightlies? We can scrub the existing releases (since no one is using them right now anyway) and re-release with semvers attached.

From what I saw when searching for standards on releasing nightlies it seems like some choose to omit the semver.

This is an issue also with `forc-binaries` because the releases don't include the semver, but that's unrelated.

mitchmindtree: > Hmm, the tarball names for nightlies actually don't include the semver

Ahh In that case feel free to ignore this suggestion!

bingcicle: >Do we want to include the semver in the releases for nightlies?

I still think this is a something to consider though I'm not sure if this matters. I guess it would be nice if we include the semver in the tarball names? But it wouldn't be a strictly necessary thing.

mitchmindtree: I'm happy to go either way on this one :+1: Any precedent set by Rust we can follow?

src/channel.rs
```diff
@@ -31,36 +33,49 @@ pub struct Channel {
...
+        let mut elems = s.split(' ');
+        let semver_str = elems
+            .next()
+            .ok_or_else(|| anyhow!("package version string is empty"))?;
+        let date_str = elems.next();
+        let semver = semver::Version::parse(semver_str).context("failed to parse semver")?;
+        let date = match date_str {
+            None => None,
+            Some(s) => Date::parse(s, format_description!("([year]-[month]-[day])"))
+                .context("specified date is invalid")
```
mitchmindtree, line 66: I think we want to return an error here in the case that there is a `date_str` but it can't be parsed, e.g. perhaps
```rust
            Some(s) => {
                let date = Date::parse(s, format_description!("([year]-[month]-[day])"))
                    .context("specified date is invalid")?;
                Some(date)
            }
```
this way we don't silently fall back to a `None` in the case a date was intended but malformed for example.

src/channel.rs
```diff
@@ -4,10 +4,12 @@ use crate::{
     file::read_file,
     toolchain::{DistToolchainName, OfficialToolchainDescription},
 };
-use anyhow::{bail, Result};
+use anyhow::{anyhow, bail, Context, Result};
+use core::fmt;
```
mitchmindtree, line 8: You can use `fmt` via `std` too - normally you only want to use `core` directly if working with `no_std` code.

src/ops/fuelup_check.rs
```diff
@@ -21,68 +21,43 @@ use std::{
...
 
     for (name, package) in channel.pkg.into_iter() {
-        latest_versions.insert(name, (package.version.semver, package.version.date));
+        latest_versions.insert(
+            name,
+            PackageVersion {
+                semver: package.version.semver,
+                date: package.version.date,
+            },
+        );
```
mitchmindtree, line 38: Maybe this could be something like the following?
```rust
latest_versions.insert(name, package.version.clone())
```

mitchmindtree (commented):
Looking much nicer! Just a couple small suggestions

src/channel.rs
bingcicle: oops, I think i accidentally picked that from the import list. Thanks for the catch!

mitchmindtree (approved):

src/download.rs
bingcicle: Just for posterity, it seems like from browsing other projects that there's no clear standard to doing this. I decided to re-release the binaries with the semver included, as well as re-publish the channels to include the updated URLs.

IMO, having the semver included in the releases as well as the URLs could clear up confusion about which latest release a certain nightly build was last tagged to; and we at least have those version numbers to refer to when working with nightlies. It's also a piece of info that wouldn't cost much to include (slightly longer tarball name), so why not.

Will be nuking the old releases/channels that don't conform to this standard since nightlies aren't live yet.

PR #170, opened 2022-08-31
nightly specify date

### LINKED ISSUE #169: support installing `nightly` from a specific date
#167 will allow downloads of the latest nightly but not with specific dates. We should be able to do `fuelup toolchain install nightly-<date>`.

bingcicle:
Closes #169 
Follow up to #167 

Allows you to install a specific date of nightly. Specifying the `TargetTriple` is still disallowed (automatically sourced from the running system's OS).

Commands that should work:
- [x]  `fuelup toolchain install nightly-2022-08-31 `
- [x]  `fuelup default nightly-2022-08-31`
- [x] `fuelup toolchain uninstall nightly-2022-08-31`

In addition to the above,
`fuelup check` should not check updates for date-specific versions
You should still not be able to add/remove components from date specific nightlies

This is not ready for review just yet - would like to look through the code again tomorrow and add more tests.


bingcicle:
This should be good to go as well but note that this merges into #167.

mitchmindtree:
Hey looks like this might have slipped by, or didn't quite make it to "Ready for review"? Happy to review once rebased or merged with master and out of draft mode :)

bingcicle:
Yep, this was sort of left behind because of the previous PRs to fix things related to the semver changes and is very outdated - will come back to this soon

src/config.rs
```diff
@@ -37,9 +38,12 @@ impl Config {
                 .filter_map(io::Result::ok)
                 .filter(|e| {
                     e.file_type().map(|f| f.is_dir()).unwrap_or(false)
-                        && RESERVED_TOOLCHAIN_NAMES
-                            .iter()
-                            .any(|t| e.file_name().to_string_lossy().starts_with(t))
+                        && RESERVED_TOOLCHAIN_NAMES.iter().any(|t| {
+                            e.file_name().to_string_lossy()
+                                == t.to_string()
```
sdankel, line 43: nit: use `format` for readability

src/constants.rs
```diff
@@ -19,3 +19,4 @@ pub const CHANNEL_LATEST_FILE_NAME: &str = "channel-fuel-latest.toml";
 pub const CHANNEL_NIGHTLY_FILE_NAME: &str = "channel-fuel-nightly.toml";
 
 pub const DATE_FORMAT: &[FormatItem] = format_description!("[year]-[month]-[day]");
+pub const DATE_FORMAT_URL_FRIENDLY: &[FormatItem] = format_description!("[year]/[month]/[day]");
```
sdankel, line 22: nit

```suggestion
pub const DATE_FORMAT: &[FormatItem] = format_description!("YYYY-MM-DD");
pub const DATE_FORMAT_URL_FRIENDLY: &[FormatItem] = format_description!("YYYY/MM/DD");
```

src/toolchain.rs
```diff
@@ -291,15 +304,15 @@ mod tests {
...
-            // TODO: uncomment once specifying date and target is supporting
-            //   assert_eq!(
-            //       desc.name,
-            //       DistToolchainName::from_str(channel::NIGHTLY).unwrap()
-            //   );
-            //   assert_eq!(desc.date.unwrap().to_string(), DATE);
-            //   assert_eq!(desc.target.unwrap().to_string(), target);
+            let toolchain = channel::NIGHTLY.to_owned() + "-" + DATE + "-" + target;
+            assert!(OfficialToolchainDescription::from_str(&toolchain).is_err());
+            // TODO: Uncomment once target specification is supported
```
sdankel, line 309: nit: add the issue link if there is one.

src/toolchain.rs
```diff
@@ -273,12 +286,12 @@ mod tests {
 
     #[test]
     fn test_parse_nightly_date() -> Result<()> {
-        assert!(OfficialToolchainDescription::from_str(NIGHTLY_DATE).is_err());
+        let toolchain = channel::NIGHTLY.to_owned() + "-" + DATE;
```
sdankel, line 289: nit: use `format`

tests/commands.rs
```diff
@@ -49,6 +51,7 @@ fn fuelup_toolchain_install_latest() -> Result<()> {
             expect_files_exist(&toolchain_dir.path().join("bin"), ALL_BINS);
 
             let output = cfg.fuelup(&["check"]);
+            println!("{:?}", output);
```
sdankel, line 54:
```suggestion
            tracing::debug!("fuelup check output: {:?}", output);
```

tests/commands.rs
```diff
@@ -80,6 +83,91 @@ fn fuelup_toolchain_install_nightly() -> Result<()> {
...
+                expected_toolchain_name,
+                toolchain_dir.file_name().to_str().unwrap()
+            );
+            assert!(toolchain_dir.file_type().unwrap().is_dir());
+
+            expect_files_exist(&toolchain_dir.path().join("bin"), ALL_BINS);
+        }
+    })?;
+
+    Ok(())
```
sdankel, line 105: Why do these tests return an empty tuple (why do they return a result at all)?

tests/testcfg.rs
```diff
@@ -81,6 +90,29 @@ impl TestCfg {
...
+    fs::create_dir_all(&bin_dir).expect("Failed to create temporary latest toolchain bin dir");
+
+    for bin in ALL_BINS {
+        fs::File::create(&bin_dir.join(bin))?;
+    }
+
+    Ok(())
+}
+
+fn setup_settings_file(settings_path: &Path, default_toolchain: &str) -> Result<()> {
```
sdankel, line 107: nit: we could make this function append `settings.toml` rather than having to do it every time this is called.

```suggestion
fn setup_settings_file(settings_dir: &Path, default_toolchain: &str) -> Result<()> {
    let settings_path = &settings_dir.join("settings.toml");
```

src/constants.rs
bingcicle: This is part of the API within the [format_description!](https://docs.rs/time/latest/src/time/format_description/component.rs.html#78) implementation (unless I'm missing something)

src/toolchain.rs
bingcicle: Ah, please ignore this comment! Honestly I don't even remember why it was there originally now because it was quite long ago but it was meant to make tests pass prior to some (now) merged changes, which weren't ready before. Targets should be automatically generated for the host machine anyway, so this shouldn't be a blocker for specifying dates.

tests/commands.rs
bingcicle: Each test does some setup prior to executing the actual tests, and within the setup we execute `setup_toolchain` which returns a Result and may fail, because it is trying to create tmp files/directories for tests. So I had to use the `?` notation for the setup of each test (which requires that the test fn returns a `Result`).

tests/testcfg.rs
bingcicle: Ah, this is way better :+1: thanks for the suggestion!

mitchmindtree (approved):

PR #172, opened 2022-09-02
nightly ci include semver

bingcicle:
This updates the 'Publish nightly channel' workflow to publish the new URLs to the nightly binaries, which now include the semver inside their names. 

Example:
Instead of `fuel-core-nightly-2022-09-02`, it will now be `fuel-core-0.10.1-nightly-2022-09-02`.

Including the semver is better than not including it, since it would dispel confusion as to which X official release that a Y nightly release was last tagged to.

You can check it out live on gh-pages branch: https://github.com/FuelLabs/fuelup/blob/gh-pages/channel-fuel-nightly.toml. Copy+paste the URL into your browser to try downloading one of the bins

### DIFF
--- a/.github/workflows/scripts/index-versions.sh
+++ b/.github/workflows/scripts/index-versions.sh
@@ -42,10 +42,9 @@ create_pkg_in_channel() {
     esac
 
     if [ "${2}" = "nightly" ]; then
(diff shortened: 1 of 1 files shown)


mitchmindtree (approved):
UtACK

Issue #162, opened 2022-08-23
[Tracking] Support sourcing forc plugins from alternative repositories


mitchmindtree:
We'll soon need to handle retrieving forc plugins from multiple locations. This will include:

- `forc-wallet` https://github.com/FuelLabs/forc-wallet
- `forc-client` https://github.com/FuelLabs/sway/issues/2608
- `forc-explore` https://github.com/FuelLabs/sway/issues/2609

It might be useful to maintain a `forc-plugins.toml` index or something along these lines that we can use from both CI and from within `fuelup` itself so that we don't need to update multiple locations of source/tests each time we add/remove a plugin.

bingcicle:
There's a few things to keep in mind here that might make this challenging:

1. We probably need some sort of standard when it comes to naming of the binary releases - for example, the Sway repo releases do not have their semver number in the name, while `fuel-core` does. The target triples that the Sway repo uses is also different from `fuel-core` (eg. forc is `amd64` versus fuel-core's `x86_64`). I chose to follow `fuel-core` in this regard when releasing `fuelup` binaries. 

The difference in naming of the binaries is also why [index-versions.sh](https://github.com/FuelLabs/fuelup/blob/168097bdcbf7f2ddb4f863da9da86d7bc022082b/.github/workflows/scripts/index-versions.sh) was a bit of a headache to deal with when adding `nightly` support. 

This isn't a problem for `fuelup`, more so for the CI since the CI handles bundling all the URLs into the channel toml, so `fuelup` doesn't have to think about how to format the URLs based on the tarball names.

2.  Which leads me to my next point, which is that the above could prove to be a problem if we want to add support for directly installing plugins from, say, a github repo. Though this is probably something far away, since we wouldn't really need `fuelup` to be able to do this unless we want devs to be able to download external plugins written by external contributors in the Fuel ecosystem. Perhaps the solution here is to just require external plugins to be downloaded with the full URL?

Maybe what we want is not really `forc-plugins.toml`, but rather `forc-components.toml` (I see plugins as a subset of components), i.e. `forc-explore` is a plugin, `forc`, the main tool, is a component. This also helps if we eventually want to distribute the indexer as well(?). This TOML file could contain info such as component name, executables, repository url and targets:

- name: the name of the component
- executables: the plugin may have several executables downloaded eg. `forc-client`
- repository_url: the distributed package can be named something different from the repo eg. sway -> forc
[...]

PR #181, opened 2022-09-08
components toml

### LINKED ISSUE #163: Add a `Plugin` abstraction that allows us to reason about plugins with multiple commands
In #152 and #154 especially we add some hacks to check for `forc-client` in order to handle its unique behaviour of providing multiple commands.

In the future, we will likely add more plugins that support multiple commands (e.g. [`forc-edit`](https://github.com/FuelLabs/sway/issues/2369)) and we should use a proper `Plugin` type (rather than just strings) that allows us to reason about this more generally.

bingcicle:
~_WIP_~

closes #163 

- Adds `Component` and a `Plugin` abstractions to be able to read supported components and plugins from `components.toml` (which will be shared between `fuelup` usage and the CI).
- For now, this PR will only address the `fuelup` side of things - another PR will address the CI side
- In future, any components that have to be added and distributed by fuelup will have to be configured within `components.toml`. By relying on reading this TOML, fuelup will automatically know which component/plugin is supported and be able to download it.

This PR should also partially address #162 - the hope is to be able to specify a new component or plugin in this TOML and have `fuelup` and the CI automatically detect them.

### DIFF
--- a/components.toml
+++ b/components.toml
@@ -5,7 +5,6 @@ executables = ["forc"]
 repository_url = "https://github.com/FuelLabs/sway"
 targets = ["linux_amd64", "linux_arm64", "darwin_amd64", "darwin_arm64"]
 
-# This might be distributed together with `forc` above, but we should have a separate entry still.
 [component.forc-fmt]
 name = "forc-fmt"
 tarball_prefix = "forc-binaries"
--- a/components.toml
+++ b/components.toml
@@ -1,0 +1,46 @@
+[component.forc]
+name = "forc"
+tarball_prefix = "forc-binaries"
+executables = ["forc"]
+repository_url = "https://github.com/FuelLabs/sway"
+targets = ["linux_amd64", "linux_arm64", "darwin_amd64", "darwin_arm64"]
+
+# This might be distributed together with `forc` above, but we should have a separate entry still.
+[component.forc-fmt]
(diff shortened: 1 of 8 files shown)


bingcicle:
This should be good to go for reviews

bingcicle:
hmm starting to think it might better to publish `components.toml` instead. This way we can freely edit the declarations for components without having to re-release fuelup. But it means having to do another network call again.

mitchmindtree (approved):
Great stuff @bingcicle, looking way cleaner :heavy_check_mark: 

> hmm starting to think it might better to publish components.toml instead. This way we can freely edit the declarations for components without having to re-release fuelup. But it means having to do another network call again.

I think these might both be good reasons *not* to fetch `components.toml`. Having it compiled in is probably good for `fuelup`'s determinism and predictability. The extra fetch might also add extra unnecessary delay? Happy to revisit this later either way!

bingcicle:
> I think these might both be good reasons _not_ to fetch `components.toml`. Having it compiled in is probably good for `fuelup`'s determinism and predictability. The extra fetch might also add extra unnecessary delay? Happy to revisit this later either way!

I think you're right about this after some thought. I'd prefer to re-release fuelup if we wanted to add/remove downloadable components rather than depend on an additional network call. Determinism and predictability is definitely more important here.

PR #183, opened 2022-09-09
install std and core libs on installing `forc`

### LINKED ISSUE #7: Installing `forc` should also install std and core libs
When downloading binaries for a new version of `forc` should also install the std and core libs, so that no further internet access is necessary. This could be done by cloning the repo for those deps exactly as `forc` would, or through another mechanism.

bingcicle:
closes #7 

This PR ensures that if we install `forc`, `fuelup` creates a temporary project within `.fuelup` using `tempfile` crate, executes `forc init --path` and `forc check --path` on it.

Note that this also executes the `forc` that is within the toolchain, so the right version is fetched.

To test:
1. rm -rf ~/.forc
2. checkout this branch
3. execute `cargo run component add forc`
4. Check that ~/.forc exists
5. Repeat step 1 again and execute `cargo run toolchain install latest`, check step 4 again

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -2,6 +2,7 @@ use anyhow::{bail, Context, Result};
 use std::fmt;
 use std::fs::{remove_dir_all, remove_file};
 use std::path::PathBuf;
+use std::process::Command;
 use std::str::FromStr;
 use time::Date;
 use tracing::info;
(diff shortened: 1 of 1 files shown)


mitchmindtree (approved):
Nice and simple :ok_hand:

PR #186, opened 2022-09-10
feat: read from toml when publishing channels

### LINKED ISSUE #191: read from `components.toml` for channel publishes
We now have #181 merged which means we have abstractions for Components and Plugins from reading `components.toml`, which means we can specify which components/plugins are supported in a declarative manner. The CI should also use the same TOML file to publish the channels that are downloadable, so that both share the same source of truth.

bingcicle:
Closes #191 
Followup to #181 

The same `components.toml` used in #181 should be used to read publishable components, and publish them within our channels (latest/nightly) in the CI.

Note that there is a caveat here with the `latest` channel publishing - because we specifically test `forc` against `fuel-core` and get the compatible versions, while reading the list of components to publish, I had to use switch block to match fuel-core to its version, and the rest (assumed to be forc and its plugins) are matched to the forc version. This is probably not an ideal way to do this but it works for now, unless there are other compatibilities we want tested as well.


bingcicle:
sorry for the mess^ commits from the old target branch was still showing even though the branch was merged into master already. Was trying to fix it by changing target branches but it didn't work so nicely because of squashed commits.

bingcicle:
Converting to draft first while I change some implementation details. There's no need for toml-reader since it has a lot of crossover with component.rs in the repo. Creating a new dir called `ci` to unify the two.

bingcicle:
Converted to draft while prioritising forc-client addition first.

PR #187, opened 2022-09-11
feat: auto detect http proxy setting.

calldata:


bingcicle (approved):
Thanks for your contribution!

PR #193, opened 2022-09-19
fix: date format when publishing

bingcicle:
after #189, date format for nightlies is no longer `YYYY-MM-DD` but `YYYYMMDD`. Missed this change in the script.


kayagokalp (approved):
utACK

PR #198, opened 2022-09-21
refactor: deprecate latest tag

### LINKED ISSUE #197: deprecate `get_latest_tag`
Key reason is to ready the code for forc-client support, and there are some redundancies here that we can cut down on if this is refactored first (such as the constants). Otherwise, we would have to add a bunch of new constants for forc-client.

The meaning of this is incorrect since our latest version is actually what is published in our channel, i.e. fuelup component add forc should fetch the version that is published within the channel.

Its also better to stop relying on github API for this purpose.

bingcicle:
Closes #197 

For various reasons:

> Key reason is to ready the code for `forc-client` support, and there are some redundancies here that we can cut down on if this is refactored first (such as the constants). Otherwise, we would have to add a bunch of new constants for `forc-client`.

> The meaning of this is incorrect since our `latest` version is actually what is published in our channel, i.e. `fuelup component add forc` should fetch the version that is published within the channel.

> Its also better to stop relying on github API for this purpose.

note that we still have to manually fetch the tag for `fuelup` since its own versioning is not published within the channel.

### DIFF
--- a/src/download.rs
+++ b/src/download.rs
@@ -101,19 +101,33 @@ pub fn tarball_name(name: &str, version: &Version, target: &TargetTriple) -> Res
 
 pub fn get_latest_version(name: &str) -> Result<Version> {
     let handle = ureq::builder().user_agent("fuelup").build();
-    let resp = handle.get(CHANNEL_LATEST_URL).call()?;
-
     let mut data = Vec::new();
-    resp.into_reader().read_to_end(&mut data)?;
-    let tmp_dir = tempdir_in(&fuelup_dir())?;
+    if name == component::FUELUP {
+        static FUELUP_API_URL: &str =
+            "https://api.github.com/repos/FuelLabs/fuelup/releases/latest";
+        let resp = handle.get(FUELUP_API_URL).call()?;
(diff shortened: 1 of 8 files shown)


src/download.rs
```diff
@@ -94,35 +99,35 @@ pub fn tarball_name(name: &str, version: &Version, target: &TargetTriple) -> Res
...
-
-    let response: LatestReleaseApiResponse = serde_json::from_str(&String::from_utf8_lossy(&data))?;
-
-    // Given a semver version with preceding 'v' (e.g. `v1.2.3`), take the slice after 'v' (e.g. `1.2.3`).
-    let version_str = &response.tag_name["v".len()..];
-    let version = Version::parse(version_str)?;
-
-    Ok(version)
+    if name == component::FUELUP {
+        static FUELUP_API_URL: &str =
```
mitchmindtree, line 106:
```suggestion
        const FUELUP_API_URL: &str =
```

mitchmindtree (approved):
Just a minor nit, otherwise :ok_hand:

mitchmindtree (approved):

PR #201, opened 2022-09-21
ci: publish forc-client in CI

### LINKED ISSUE #185: include new package: `forc-client`
Accompanying forc-client issue: https://github.com/FuelLabs/forc-client/issues/2

We should distribute forc-client binary by including it in the channels through the CI.

Checklist:

- [ ] publish `forc-client` within the `latest` and `nightly` channel

bingcicle:
Closes #185 

Add `forc-client` to channel within the CI.

Note that in the `latest` channel, we're hardcoding versions for now. 

We probably want the same `forc-client` that is used to run tests here, since `forc-client` is fetched as a dep in the tests. However, forc-client version is also fixed there. Perhaps we also need some sort of detection mechanism to run tests everytime a new forc-client is released.

### DIFF
--- a/.github/workflows/publish-nightly-channel.yml
+++ b/.github/workflows/publish-nightly-channel.yml
@@ -30,6 +30,7 @@ jobs:
             echo -e "published_by = \"https://github.com/FuelLabs/fuelup/actions/runs/${GITHUB_RUN_ID}\"\n" >> $CHANNEL_TOML
 
             ./.github/workflows/scripts/append_package_to_channel.sh forc nightly $CHANNEL_TOML
+            ./.github/workflows/scripts/append_package_to_channel.sh forc-client nightly $CHANNEL_TOML
             ./.github/workflows/scripts/append_package_to_channel.sh fuel-core nightly $CHANNEL_TOML
 
             # Remove extra newline at the end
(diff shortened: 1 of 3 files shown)


mitchmindtree (approved):
> Perhaps we also need some sort of detection mechanism to run tests everytime a new forc-client is released.

True, that makes sense, though I guess the matrix of package combinations we need to test might explode if we keep adding more into the Sway integration tests? I wonder if for "latest" we're better off just selecting the versions that are currently used in the Sway integration tests :thinking: I guess the problem there is that the fuel-core (and now forc-deploy) can often be multiple versions behind.

bingcicle:
>  though I guess the matrix of package combinations we need to test might explode if we keep adding more into the Sway integration tests?

That's my main concern, unless there's only a select number of components/plugins that we want to ensure compatibility. For example, it's not clear if `forc-wallet` would require compatibility tests like `forc-client` does (since forc-client directly interacts with the node).

Perhaps when it gets too complex, we would already be releasing the `latest` toolchain on a time-based schedule instead of being on a on-demand based schedule (where we try to test asap), in that case it wouldn't be much of an issue. Will open an issue for this though.

PR #204, opened 2022-09-22
feat: extract src/component.rs into its own lib

bingcicle:
Precursor to #186 - extracting `src/component.rs ` as its own library so that we can use a Rust script to generate the channels. Note that this PR was cherrypicked from a commit in that PR since this is technically not related to the CI change.


Cargo.toml
```diff
@@ -20,6 +20,7 @@ path = "src/main.rs"
 [dependencies]
 anyhow = "1"
 clap = { version = "3.2", features = ["cargo", "derive", "env"] }
+component = { path = "component" }
```
mitchmindtree, line 23: This will need to specify the version as well if we wish to publish future `fuelup` versions to crates.io.

component/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "component"
+version = "0.0.0"
+edition = "2021"
```
mitchmindtree, line 4: We'll need the rest of the metadata if we're to publish future `fuelup` versions to crates.io.

mitchmindtree (commented):
I don't recall if we publish this to crates.io or not? If so, we should address these two suggestions, otherwise maybe we should explicitly add `publish = false` to both crates?

bingcicle:
> I don't recall if we publish this to crates.io or not? If so, we should address these two suggestions, otherwise maybe we should explicitly add `publish = false` to both crates?

Ah yes, no this shouldn't be published. Thanks for the catch - I think I missed this when I cherry picked it out of the PR

mitchmindtree (approved):

mitchmindtree (approved):

PR #202, opened 2022-09-21
feat: allow fuelup to download forc-client

bingcicle:
closes #199 

General support for forc-client. `fuelup` should be able to run all the commands successfully.

PR #206, opened 2022-09-23
release: v0.7.0

bingcicle:
Contains `forc-client` support, and contains changes that bumps `src/component.rs` as its standalone crate.

Could be worth waiting for a new `forc` release (since the current one still contains `forc-client`.

Depends on:
-  [x] #204 
- [x] #201 
- [x] #202


eureka-cpu (approved):
🚀

PR #210, opened 2022-09-27
ci: script to read from `components.toml` when building channels

### LINKED ISSUE #207: rewrite script to build channels
Currently we have a barebones [script](https://github.com/FuelLabs/fuelup/blob/ccf05d29eed87018b4a3196cdcc34c18f74902dc/.github/workflows/scripts/append_package_to_channel.sh) to build our channels. Since we have the `Component` abstraction and `components.toml` thanks to #181  I think it's preferable we rewrite the script at some point in Rust to ensure reliability as we add/remove/modify components within the TOML.

bingcicle:
Reopened from #186 
Closes https://github.com/FuelLabs/fuelup/issues/191
closes #207 
Followup to https://github.com/FuelLabs/fuelup/pull/181

We have `append_package_to_channel.sh` which we use to create channels, but this gets harder to maintain and adapt as we add more components to distribute.

The same components.toml used in https://github.com/FuelLabs/fuelup/pull/181 should be used to read publishable components, and publish them within our channels (latest/nightly) in the CI. Then the original shell script can be deprecated.

This is better because:
- `components.toml` is declarative in nature - this means that once this PR is in, if we want to support/deprecate a component, we just need to edit that TOML file.
- `component.rs` has been extracted as a lib which means we can leverage its abstractions to create a channel toml in more reliable and testable manner.


bingcicle:
This should be good to go for reviews - the use case primarily is centred around usage for the CI. 

I manually ran the the nightly publish CI once here: https://github.com/FuelLabs/fuelup/actions/runs/3132165597 and the resulting channel was here (has since been replaced by the actual nightly though): https://github.com/FuelLabs/fuelup/blob/88bb91b8df94cd6ca9e5c9de072559071dab19c6/channel-fuel-nightly.toml

JoshuaBatty (approved):
utACK

PR #215, opened 2022-09-28
refactor: deprecate usage of consts

### LINKED ISSUE #205: deprecate consts used for components
While adding forc-client support to fuelup I noticed there are some consts still being used within the codebase, mainly to handle components (tarball names, component names come to mind). These should be ideally deprecated as we move towards a more declarative way to supporting components as we see in `components.toml`.

bingcicle:
closes #205 

Purpose of this PR is to reduce the reliance of const strings to check if an input is a component or not, and instead rely on reading `components.toml`.


kayagokalp (approved):
lgtm

PR #220, opened 2022-10-01
bug: fix download panick when fuelup is not updated

### LINKED ISSUE #219: installing `latest` toolchain fails if not on latest fuelup
### Problem

When adding support for `forc-wallet`, running the installation command panicks because the old `fuelup` version does not know how to handle the newly published `forc-wallet` component until we upgrade it to the latest version (0.8.0).

```console
$ fuelup toolchain install latest
thread 'main' panicked at 'Could not create DownloadCfg from a package parsed in latest channel: Unrecognized component: forc-wallet', src/channel.rs:74:22
```

Ideally we should continue with installation but warn about any components that can't be downloaded.

### Steps

1. Be on `fuelup <0.8.0`.
2. Run `fuelup toolchain install latest`

### Possible Solution(s)

while building and collecting download configs, we should filter out any Errs and ignore them, and download the rest.

### Notes

_No response_

### Fuelup version

```console
0.7.1
```

### Installed components

```console
nil
```

bingcicle:
closes #219 

Filter `DownloadCfg` creation by taking only the configs that are `Ok`. The bug was a result of what was described within the issue, a mismatch between what the `fuelup` binary knows how to handle and the published channel. Upon `0.8.0` release, `forc-wallet` was published within the channel, but v0.7.1 does not yet know that it does, until you update to v0.8.0.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -258,7 +258,7 @@ dependencies = [
 
 [[package]]
 name = "fuelup"
-version = "0.7.1"
+version = "0.8.0"
 dependencies = [
  "anyhow",
  "clap",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fuelup"
-version = "0.7.1"
+version = "0.8.0"
 authors = ["Fuel Labs <contact@fuel.sh>"]
 edition = "2021"
 homepage = "https://fuel.network/"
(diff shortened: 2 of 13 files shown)


mitchmindtree (commented):
Looks good, though perhaps we should emit a warning if we detect a package for a channel that cannot be handled and suggest updating `fuelup` before trying again?

mitchmindtree (approved):
:ok_hand:

PR #224, opened 2022-10-03
remove publish = true for `forc-client`

bingcicle:
Remove forc-client support.


Braqzen (approved):

mitchmindtree (approved):
Just to clarify, we're removing this as it's the default behaviour right?

bingcicle:
> Just to clarify, we're removing this as it's the default behaviour right?

Removing this since `fuelup` shouldn't be able to download/use forc-client as its own component (the `publish = true` kv pair distinguishes it as a separate component that has been published), but it still counts as part of `forc`.

Issue #230, opened 2022-10-04
tmp dirs not being cleaned up


bingcicle:
Noticed this while working on trying to implement `fuelup update`, but tmp dirs weren't getting cleaned up properly after `fuelup toolchain install` command. This is polluting the `.fuelup` directory.

Issue #208, opened 2022-09-24
`fuelup update`


bingcicle:
Now that we have 2 channels we should be able to do `fuelup update` to update both `latest` and `nightly` toolchains.  Basically running both `fuelup toolchain install latest` and `fuelup toolchain install nightly` provided the toolchains were already installed.

Issue #190, opened 2022-09-15
correctly implement nightly versioning to comply with semver


bingcicle:
The previous nightlies did not comply with semver spec #10: https://semver.org/#spec-item-10

> Build metadata MAY be denoted by appending a plus sign and a series of dot separated identifiers immediately following the patch or pre-release version

Previously, our `nightly` tag was appended to, for example, `v1.0.0`, to mean that it was a `nightly` build of the released version `v1.0.0`. As it turns out, according to the specification, the correct way of interpreting `v1.0.0-nightly` is that it was a _pre-release_ - and hence `v1.0.0-nightly` < `v1.0.0`.

The versions should actually be in the format `0.24.3+nightly.20220915.abcdefg`, where:
- 0.24.3 = version core
- nightly = build identifier
- 20220915 = build identifier (date)
- abcdefg = build identifier (git commit hash)

PRs have been made in the sway-nightly-binaries repo to correct this implementation error.

PR #232, opened 2022-10-05
refactor: use new subdir /tmp for std-lib checkouts

bingcicle:
related: #230 #231 

starts a new Forc project within `.fuelup/tmp` sub directory instead of within `.fuelup`.

Reason:

there isn't a way to guarantee that the temporary directories created while fetching Forc dependencies (via init-ing and checking a Forc project) are cleaned up, especially if we kill the download before it finishes.

This is reproduced by installing forc (either by `fuelup toolchain install latest` or `fuelup component add forc`), getting to the part where we start fetching forc dependencies (terminal says `Fetching core forc dependencies`) and stopping/killing the process. We are left behind with a `temp-projectABCD` directory which is left uncleaned.
 
Solution can be to save all these temp projects within a sub directory `tmp`, and we try to clean up that directory instead, perhaps through another command.

### DIFF
--- a/component/src/lib.rs
+++ b/component/src/lib.rs
@@ -73,7 +73,7 @@ impl Components {
         Ok(components)
     }
 
-    pub fn contains(name: &str) -> bool {
+    pub fn contains_published(name: &str) -> bool {
         Self::collect_publishables()
             .expect("Failed to collect publishable components")
             .iter()
--- a/components.toml
+++ b/components.toml
@@ -37,7 +37,6 @@ is_plugin = true
 executables = ["forc-deploy", "forc-run"]
 repository_name = "forc-client"
 targets = [ "aarch64-unknown-linux-gnu", "x86_64-unknown-linux-gnu", "aarch64-apple-darwin", "x86_64-apple-darwin" ]
-publish = true
 
 [component.fuel-core]
 name = "fuel-core"
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -5,6 +5,7 @@ use crate::{
     toolchain::{DistToolchainName, OfficialToolchainDescription},
 };
 use anyhow::{bail, Result};
+use component::Components;
 use semver::Version;
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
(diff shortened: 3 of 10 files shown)


Braqzen (commented):
I'm unsure of the bigger picture but isn't there another PR which is supposed to remove touching the disk / create tmp files and instead do something in memory?

bingcicle:
> I'm unsure of the bigger picture but isn't there another PR which is supposed to remove touching the disk / create tmp files and instead do something in memory?

Yup - though that was for the channel toml, and this will be to address fetching forc dependencies that we do in the background when installing the toolchain!

In #231 the entire toml file was previously being saved for 2 reasons: 

1) to use to compare versions - this was prior to having proper semver because of incorrect implementation on my part - see issue #190 and the accompanying PR #189, and 
2) it was convenient to use the same function `download_file` which we used to download the actual binaries anyway. 

The first has since been fixed by the PR linked above and 2) was just addressed in #231 where we use a different function where we do not persist the data as a file.

Now, for this PR, what installing `forc` does is that it checks out the std-lib to `~/.forc` by doing `forc init --path <path>` and `forc check --path <path>`, and this would've been done in a [tempdir](https://docs.rs/tempfile/latest/tempfile/fn.tempdir.html). However, there isn't a way to ensure that the destructors run (user could cancel/kill the process before completion, for example). This PR is supposed to use a subdir within `~/.fuelup` so that at least the base `.fuelup` directory isn't polluted, and that we have a dedicated `tmp` dir.

sdankel (approved):
> and we try to clean up that directory instead, perhaps through another command.

It makes sense that it'll be easier to clean if these files are in a separate subdirectory, but I'm curious what command will do the cleanup?

bingcicle:
> It makes sense that it'll be easier to clean if these files are in a separate subdirectory, but I'm curious what command will do the cleanup?

For `rustup` there is an implicit cleanup done on every `rustup update`. I think going in line with what they're doing makes sense here, to do a cleanup on `fuelup update` (yet to be implemented, #208) or `fuelup self update`, and printing/logging a line to let the developer know that this is happening. I don't think it would be good for DevEx to have an explicit `fuelup clean` command to delete the contents of `/tmp` (though that may be an option as well) since it may not be immediately clear what the command would be for. Open to suggestions on this!

PR #238, opened 2022-10-20
feat: use update hashes

### LINKED ISSUE #69: `fuelup toolchain install latest` should not fetch / reinstall versions that a user already has fetched / installed

bingcicle:
closes #69 
unblocks #208 

- saves the sha256 hash of the channel to a file with file name of the toolchain eg. `latest-x86-apple-darwin` to a new directory, `~/.fuelup/hashes`
- we use this hash to avoid re-downloading a toolchain if it already exists.
- the hash is saved ONLY if the entire toolchain is installed (skipped if there are partial installations)
- the hash is deleted upon uninstallation of a toolchain i.e. `fuelup toolchain uninstall latest`

This unblocks #208 since we want to check for hashes before trying to update `latest` and `nightly`.

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -35,7 +35,7 @@ pub struct Package {
 }
 
 impl Channel {
-    pub fn from_dist_channel(desc: &OfficialToolchainDescription) -> Result<Self> {
+    pub fn from_dist_channel(desc: &OfficialToolchainDescription) -> Result<(Self, String)> {
         let channel_file_name = match desc.name {
             DistToolchainName::Latest => CHANNEL_LATEST_FILE_NAME,
             DistToolchainName::Nightly => CHANNEL_NIGHTLY_FILE_NAME,
@@ -47,7 +47,8 @@ impl Channel {
             Err(_) => bail!("Could not read {}", &channel_url),
         };
 
-        Self::from_toml(&toml)
+        let actual_hash = format!("{:x}", hasher.finalize());
+        Ok((Self::from_toml(&toml)?, actual_hash))
     }
 
     pub fn from_toml(toml: &str) -> Result<Self> {
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -38,6 +38,7 @@ pub struct Package {
 }
 
 impl Channel {
+    /// The returned `String` is a sha256 hash of the downloaded toolchain TOML bytes.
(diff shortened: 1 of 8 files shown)


src/config.rs
```diff
@@ -1,24 +1,54 @@
...
+    pub(crate) fn hash_matches(
+        &self,
+        description: &OfficialToolchainDescription,
+        hash: &str,
+    ) -> bool {
+        let hash_path = self.hashes_dir.join(description.to_string());
+
+        match read_file("hash", &hash_path) {
+            Ok(h) => h == hash,
+            _ => false,
```
mitchmindtree, line 38: Rather than treating all types of I/O errors as a hash mismatch, we should only return `false` if the I/O error indicates that the file doesn't exist, or if `h != hash`, otherwise we should probably return the error.

src/channel.rs
```diff
@@ -38,7 +38,7 @@ pub struct Package {
 }
 
 impl Channel {
-    pub fn from_dist_channel(desc: &OfficialToolchainDescription) -> Result<Self> {
+    pub fn from_dist_channel(desc: &OfficialToolchainDescription) -> Result<(Self, String)> {
```
mitchmindtree, line 41:
```suggestion
    /// The returned `String` is a sha256 hash of the downloaded toolchain TOML bytes.
    pub fn from_dist_channel(desc: &OfficialToolchainDescription) -> Result<(Self, String)> {
```

src/config.rs
bingcicle: The initial thought was that the only time that a toolchain should skip installation was if the hashes matches and to re-install for every other case, but I guess I shouldn't be lazy and include explicitly the situations where this happens :p

mitchmindtree (dismissed):

mitchmindtree:
Looking good, just clippy nits

mitchmindtree (approved):

PR #250, opened 2022-10-26
fix: print output with fuelup show and replace `println` with  `tracing`

### LINKED ISSUE #249: `fuelup show` output is malformed
image

As seen in the screenshot, probably due to the way errors are handled.

bingcicle:
closes #249 

Fixed output: 
image

Also replaced calls of `println` with `tracing::info` or `tracing::error`.


bingcicle:
There seems to be an issue with grabbing `forc-lsp` version but it's unrelated with the original issue of malformed output. Can't execute `forc-lsp --version` for some reason.

PR #253, opened 2022-10-31
bug: replace `rename` usage with remove+copy usage

### LINKED ISSUE #252: `fuelup self update` fails to `rename` fuelup since 'from' and 'to' are on different filesystems
As reported by @segfault-magnet:

```sh
❯ fuelup self update
Fetching binary from https://github.com/FuelLabs/fuelup/releases/download/v0.10.1/fuelup-0.10.1-x86_64-unknown-linux-gnu.tar.gz
Downloading component fuelup without verifying checksum
Unpacking and moving fuelup to /tmp/.tmpEeIk4W
Moving /tmp/.tmpEeIk4W/fuelup to /home/segfault_magnet/.fuelup/bin/fuelup
Failed to replace old fuelup with new fuelup: Invalid cross-device link (os error 18). Attempting to restore backup fuelup.
fuelup failed to update itself: Old fuelup restored because something went wrong replacing old fuelup with new fuelup.

❯ fuelup --version
fuelup 0.10.0
```

As the docs for `std::fs::rename` explains ([link](https://doc.rust-lang.org/std/fs/fn.rename.html#errors)), one of the failure conditions is if `from` and `to` are on different filesystems, and that seems to be the case for manjaro.

bingcicle:
closes #252 

In order to make `copy` work without running into ETXTBSY, we should first unlink the binary first.

### DIFF
--- a/src/ops/fuelup_self.rs
+++ b/src/ops/fuelup_self.rs
@@ -6,7 +6,7 @@ use tracing::{error, info};
 
 use crate::{
     download::{download_file_and_unpack, unpack_bins, DownloadCfg},
-    path::{fuelup_bin, fuelup_bin_dir},
+    path::fuelup_bin,
     target_triple::TargetTriple,
 };
 
--- a/src/ops/fuelup_self.rs
+++ b/src/ops/fuelup_self.rs
@@ -1,12 +1,12 @@
 use anyhow::{bail, Result};
 use component;
 use std::{fs, path::Path};
-use tempfile::tempdir;
+use tempfile;
 use tracing::{error, info};
 
 use crate::{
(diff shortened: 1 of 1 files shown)


kayagokalp (approved):
utACK

bingcicle:
@kayagokalp Tests here are flaky at times and run into file i/o issues probably due to `cargo test`'s parallelization, so it's probably due to buggy tests/test setup on my part. Usually re-running the job fixes it, but we should definitely look into it at some point!

kayagokalp (approved):
utACK, I believe it should do the trick but unfortunately I do not have a manjuro setup to test

PR #259, opened 2022-11-06
bug: uninstall+default bug

### LINKED ISSUE #258: uninstalling a custom toolchain does not re-set a default toolchain
### Problem

As described above in title.

### Steps

Assuming starting with another toolchain already installed,

1. fuelup toolchain new my-toolchain
2. fuelup toolchain uninstall my-toolchain
3. fuelup default

`fuelup default` still shows the uninstalled `my-toolchain` as default.

### Possible Solution(s)

We should default back to the `latest` or `nightly` toolchain if possible in that order, and if they don't exist exist we should switch to the first custom toolchain in alphabetical order.

### Notes

_No response_

### Fuelup version

```console
10.2
```

### Installed components

```console
-
```

bingcicle:
closes #258 

When uninstalling a toolchain, if the toolchain is the currently active toolchain, `fuelup` should re-set a default.

This is done by doing `list_toolchains()` and selecting, in order of priority, 

an official toolchain (latest, nightly, in order of priority), or
a custom toolchain, arranged alphabetically

This PR also includes a refactor on the `default` command to split it into commands/ops.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -5,8 +5,8 @@ use anyhow::{bail, Result};
 use std::io::{self, ErrorKind};
 
 use crate::file::write_file;
+use crate::fmt::format_toolchain_with_target;
 use crate::path::{ensure_dir_exists, hashes_dir, toolchains_dir};
-use crate::target_triple::TargetTriple;
 use crate::toolchain::{OfficialToolchainDescription, RESERVED_TOOLCHAIN_NAMES};
 
 pub struct Config {
@@ -62,14 +62,10 @@ impl Config {
                 .filter(|e| e.file_type().map(|f| f.is_dir()).unwrap_or(false))
             {
                 let toolchain = dir_entry.file_name().to_string_lossy().to_string();
-                if RESERVED_TOOLCHAIN_NAMES.iter().any(|t| {
-                    toolchain
-                        == format!(
-                            "{}-{}",
-                            t,
-                            &TargetTriple::from_host().unwrap_or_default().to_string()
-                        )
-                }) {
+                if RESERVED_TOOLCHAIN_NAMES
+                    .iter()
+                    .any(|t| toolchain == format_toolchain_with_target(&t))
+                {
                     toolchains.push(toolchain)
                 } else {
                     custom_toolchains.push(toolchain)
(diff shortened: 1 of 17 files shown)


src/ops/fuelup_default.rs
```diff
@@ -0,0 +1,40 @@
...
+            info!("{} (default)", current_toolchain.name);
+            return Ok(());
+        }
+    };
+
+    let mut new_default = Toolchain::from_path(&toolchain)?;
+
+    if let Ok(description) = OfficialToolchainDescription::from_str(&toolchain) {
+        new_default = Toolchain::from_path(&description.to_string())?;
+    }
```
mitchmindtree, line 26: Here it looks like we are unnecessarily loading `new_default` twice from different paths in the case that `OfficialToolchainDescription` is `Ok`. Perhaps this could be:

```rust
let new_default = match OfficialToolchainDescription::from_str(&toolchain) {
    Ok(desc) => Toolchain::from_path(&description.to_string())?,
    Err(_) => Toolchain::from_path(&toolchain)?,
};
```

I realise this was copy pasted and previously overlooked, happy to leave for a future PR if preferred.

src/config.rs
```diff
@@ -54,13 +54,32 @@ impl Config {
...
-                .map(|e| e.file_name().into_string().ok().unwrap_or_default())
-                .collect();
+            {
+                let toolchain = dir_entry.file_name().to_string_lossy().to_string();
+                if RESERVED_TOOLCHAIN_NAMES.iter().any(|t| {
+                    toolchain
+                        == format!(
+                            "{}-{}",
+                            t,
+                            &TargetTriple::from_host().unwrap_or_default().to_string()
```
mitchmindtree, line 70: Do we have a function we can re-use for this toolchain-target-triple formatted string?

mitchmindtree (approved):
Left a couple of optional suggestions, otherwise seems fine :+1:

src/ops/fuelup_default.rs
bingcicle: Ah, this is not unnecessarily loaded. The reason is if you run `fuelup default latest`, the `latest` string must be appended with the relevant `TargetTriple` generated from the host machine, which is why `new_default` is mut and we check if we can form an `OfficialToolchainDescription` from the user input string.

I'll add in a short comment to explain this.

bingcicle: I'll create a smaller PR for this!

src/config.rs
bingcicle: Hmm, we only have 2 `format!`s here and both are used only within `config.rs`, so I didn't see the need for a function. I could create one within the same file.

mitchmindtree: The idea of using a function is so that if the format changes in the future for whatever reason, we only have to change it in one location, and avoid running the risk of forgetting to update one of the locations.

src/ops/fuelup_default.rs
bingcicle: I'll just add it to this since I'm making the change below

mitchmindtree (approved):
:ok_hand:

PR #270, opened 2022-11-10
feat: beta-2 toolchain

bingcicle:
supports installation of `beta-2` toolchain: `fuelup toolchain install beta-2`.


kayagokalp (approved):
Hey @bingcicle, tested installing with `fuelup toolchain install beta-2` and it looks good to me! I have a question though, it installs the following versions:

```console
Installed:
- forc 0.31.0
- forc-explore 0.28.1
- forc-wallet 0.1.2
- fuel-core 0.14.1
```

We do have 0.31.1 released for sway should it install that version instead? 🤔

bingcicle:
Channel updated to use `forc 0.31.1` instead. Should be good to go :+1:

PR #279, opened 2022-11-17
do fs::remove_file before fs::copy in case of running binaries

### LINKED ISSUE #278: `fuelup update` corrupts running binary on M1 Mac
### Problem

With an already running executable (fuel-core), running `fuelup update` has unexpected behaviors on different systems:
- On x86_64 Linux, the executable is not updated.
- On an M1 Mac (aarch64-apple-darwin), the executable is corrupted.

Attempting to run fuel-core after running `fuelup update` while a fuel-core instance is running results in the below output:

```sh
fuel-core --version
zsh: killed     fuel-core --version
```

### Steps

1. `fuel-core` (first run `fuel-core`)
2. `fuelup update`
3. `fuel-core --version` (this should fail)

If you already have the latest `fuel-core`, manually install an older `fuel-core` and move it into the `latest` toolchain:
1. `fuelup toolchain new bugfix-toolchain`
2. `fuelup component add fuel-core@0.14.0`
3. `mv ~/.fuelup/toolchains/bugfix-toolchain/bin/fuel-core ~/.fuelup/toolchains/latest-<TARGET>/bin`
4. `fuelup default latest`

Then run the steps above.

### Possible Solution(s)

_No response_

### Notes

_No response_

### Fuelup version

```console
`v0.12.1`
```

### Installed components

```console
not required
```

bingcicle:
Closes #278 

This was also the solution to a somewhat related problem we had prior, related to ETXTBSY bug: https://github.com/FuelLabs/fuelup/pull/253

The solution is to unlink (`fs::remove_file`) the running executable prior to trying to do an `fs::copy`.

Noticed that there's a `dst_dir` in the next line which could be confusing - this should probably be fixed in a separate refactor PR.

### DIFF
--- a/src/download.rs
+++ b/src/download.rs
@@ -303,6 +303,9 @@ pub fn unpack_bins(dir: &Path, dst_dir: &Path) -> Result<Vec<PathBuf>> {
                     &bin_file_name.to_string_lossy(),
                     dir.display()
                 );
+                if dir.join(&bin_file_name).exists() {
+                    fs::remove_file(dir.join(&bin_file_name))?;
+                }
                 fs::copy(bin_file.path(), dir.join(&bin_file_name))?;
                 downloaded.push(dst_dir.join(bin_file_name));
             }
--- a/src/download.rs
+++ b/src/download.rs
@@ -303,10 +303,12 @@ pub fn unpack_bins(dir: &Path, dst_dir: &Path) -> Result<Vec<PathBuf>> {
                     &bin_file_name.to_string_lossy(),
                     dir.display()
                 );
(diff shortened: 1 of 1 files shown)


kayagokalp (approved):
Tested with the procudure we followed offline, just listing here for documentation purposes:

1. Create a custom toolchain with (any) older version of fuel-core
2. Move the fuel-core binary to latest toolchain manually
3. Switch to latest toolchain and spin up the fuel-core
4. While it is running run fuelup update
5. Try to run again.

Before this PR, this was corrupting the binary and we couldn't run it on a mac os and with this, it works ✅

PR #281, opened 2022-11-22
feat: add fuel-indexer

bingcicle:


deekerno (approved):
❤️

ra0x3:
@bingcicle A few questions
- Is the indexer binary installed automatically when I install `fuelup`?
  - I imagine it's installed in the same location as `forc` (for example)?
- If I'm open `fuelup` and I want to change which version of the indexer binary I'm using, is that possible? If so, how would I do that?
- If I'm on a prior version of `fuelup` (0.12) would I `fuelup self update` to upgrade to the `fuelup` version that supports the indexer binary? (when that version is released, anyway)

PR #282, opened 2022-11-23
feat: handle no such file error

bingcicle:
The most common error I believe that is encountered is the `No such file or directory (os error 2)` when trying to execute a component from a toolchain which does not have the component installed. This is because fuelup tries to directly execute the executable from path, and if not found, it means that the file does not exist.

Here we handle `ErrorKind::NotFound` in a custom way to give the users more info on what happened in the context of fuelup and components.

### DIFF
--- a/src/proxy_cli.rs
+++ b/src/proxy_cli.rs
@@ -1,5 +1,6 @@
 use anyhow::Result;
 use std::ffi::OsString;
+use std::io::{Error, ErrorKind};
 use std::os::unix::prelude::CommandExt;
 use std::process::{Command, ExitCode, Stdio};
 use std::{env, io};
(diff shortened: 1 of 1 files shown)


kayagokalp (commented):
Not sure if this is again happening due to MacOs but this seems to have no effect on my system I still got `No such file..` 🤔

bingcicle:
@kayagokalp To test this, you'd have to use this version of `fuelup` to execute the other binaries since they are called through fuelup. You'd also need to re-link the binaries to `fuelup`, otherwise this new piece of code wouldn't be available in the older version (since the binaries in `/.fuelup/bin` are just fake `fuelup` binaries that call the actual binaries!). AFAIK I don't think there is a way to retroactively fix this :/

To test:

1. `mv target/debug/fuelup ~/.fuelup/bin` - ensure that we are using this fuelup instead of release fuelup
2. `fuelup toolchain new new-toolchain`
3. `forc` - executing `forc` here should still show the old error ("No such file...") like you mentioned. We have to re-link the fake binary first.
4. `fuelup component add forc` - relinks `forc` to `fuelup`. This ensures the fake `forc` has the latest error matching code block.
5. `fuelup component remove forc` - we delete the real `forc` here in the toolchains directory.
6. `forc` - this should show the new error, and will do so for future new custom toolchains.

kayagokalp (approved):
Thanks for the great testing steps @bingcicle. Everything works! 🚀 🚢

PR #292, opened 2022-11-28
feat: component list

### LINKED ISSUE #264: Proposal to add the ability for `fuelup` to list components that you have the ability to install on your current toolchain

I've been finding myself wanting the ability to write `fuelup component list` to see a list of possible components that I could install (and which version is `latest`).

### Solution you'd like

Currently my flow is to use `fuelup show` to get the current list of components that I _don't_ have installed, then `fuelup component add <name>` based on the list from the previous command. But remembering that flow feels a little unintuitive to me because `fuelup show` is semantically different than `fuelup component`, and also the description for `fuelup show` doesn't list "component": "Show the active and installed toolchains, as well as the host and fuelup home"

### Notes

Just a thought 😄

bingcicle:
closes #264

There's some extra stuff that I added/changed to the spec mentioned in the issue:
- Changed naming of __available__ to __installable__, I believe this is a more accurate naming than __available__ (something can be installed and yet still be available)
- Continue the pattern of listing `forc` plugins that are distributed together with `forc`, underneath the `forc` entry. This way we do not confuse users that plugins like `forc-fmt` are installable on their own, rather that they come packaged together with `forc`.
- Added the currently active toolchain name in the header so that the user knows the context in which they are working under.

Sample output on `latest` (with failure on `fuel-indexer` included since it's not installable yet, but is available on my development branch):
image

Sample output on a custom toolchain. Notice that `forc-wallet` is not on the latest, and we have an installable component `fuel-indexer` available:

image


src/ops/fuelup_component/list.rs
```diff
@@ -0,0 +1,101 @@
...
+
+            installed_components_summary.push_str(&format_installed_component_info(
+                &component.name,
+                current_version,
+                &version_info,
+            ));
+
+            if component.name == component::FORC {
+                installed_components_summary
+                    .push_str(&format_forc_default_plugins(component.executables))
```
mitchmindtree, line 81: Are we sure we want to have a special case for the default plugins rather than listing them like the rest? Is it possible for a user to have a different version of `forc` to say `forc-lsp` if they're using a custom toolchain?

mitchmindtree (commented):
Nice, this will be very handy :+1:

src/ops/fuelup_component/list.rs
bingcicle: Hmm, this is tricky because of the way we release forc binaries - they're all packaged together in one tarball in [this step](https://github.com/FuelLabs/sway/blob/12aff7873345daf8930affa472c222ffb7906c5d/.github/workflows/ci.yml#L626-L629). So it's not possible to install `forc-lsp` separately from `forc`, and listing them in the way I did was to emphasize that it's impossible to do that. Otherwise, we might have people trying to do `fuelup component add forc-lsp` when the command should be `fuelup component add forc`.

It's hard to get a different version of `forc` vs `forc-lsp`, but i guess its possible if the user manually moves their binaries around.

To address this, we could update this behaviour to list the versions of each plugin tagged to `forc` as well, similar to how `fuelup show` does it. But perhaps a longer term solution might be one of the below, to be able to install, say `forc-lsp`, separately:

1) we re-work the way we package the `forc` binaries (I've been wondering recently why we didn't do what [`tracing` does with their releases](https://github.com/tokio-rs/tracing/releases)), or
2)  re-work unpacking the tarball to keep only the binary we want to specifically install and delete the others (though this would probably over-complicate the process for little to no gain, so I don't really like this)

mitchmindtree: > So it's not possible to install forc-lsp separately from forc, and listing them in the way I did was to emphasize that it's impossible to do that. Otherwise, we might have people trying to do fuelup component add forc-lsp when the command should be fuelup component add forc.

Ahhh if that's the case, let's not worry about it - this seems fine :+1: It would be a very niche need anyway, and there are trivial workarounds (like installing forc-lsp via cargo instead). We can reconsider splitting up `forc` if we get a motivating issue, but otherwise this seems fine.

mitchmindtree (approved):

src/ops/fuelup_component/list.rs
bingcicle: I think I will create a separate issue to print the versions of the plugins themselves as well, because it is indeed an oversight if the actual executables are moved around by the user and are different from the `forc` version.

PR #294, opened 2022-11-30
feat: warn existing executables

### LINKED ISSUE #290: Upon installation of a toolchain or component, warn the user if any executables already existing on their `PATH`
It is totally valid for a user to have two installations of a particular tool. E.g. a user might setup their path like so:
```
PATH="~/.cargo/bin:~/.fuelup/bin"
```
This allows them to use `cargo` to temporarily install tools during development (e.g. `cargo install --path forc`) and then remove them again later `cargo uninstall forc` in a manner that falls back to their `fuelup` installation.

However, a lot of the time (especially for less-experienced devs) having multiple installations of a tool on a `PATH` can just be confusing.

During installation, we should make it clear if `fuelup` is installing a tool that is already installed on the user's system. We can check whether each of the executables already exist on the user's system by checking and searching the user's `PATH` (see `forc`'s plugin support for an example of searching the `PATH`).

If a duplicate is found we should clarify the following when installing:

- That a duplicate exists.
- Where the duplicate is located (its path).
- Which version (i.e. the existing version or the version being installed by `fuelup`) will take priority based on the search order of their `PATH`.

The final point might be trickier depending on the complexity of the user's `PATH`, but would likely be very valuable information for many users.

We could re-warn on `fuelup update` too.

---

Later on, we can consider adding heuristics for suggestions on how to resolve conflicts based on where the duplicates are and the `PATH` order, e.g. "`fuelup` detected a version of `forc` installed via `cargo` that will take precedece - consider uninstalling it with `cargo uninstall forc`".

bingcicle:
closes #290 

This check is invoked upon `install` or `update` commands.

Example output:

image

This check will warn when:

1) your fuel related executable exists but is not in `.fuelup/bin`,
2) your fuel related executable exists, is not in `.fuelup/bin` and instead is in `.cargo/bin`, urging the user to do `cargo uninstall` when `fuelup` is about to do an install or update.
3) your fuel related executable exists, is both in `.fuelup/bin` and in `.cargo/bin`, warning that the version in `.fuelup/bin` is overshadowed and urging the user to do `cargo uninstall`.

A fourth variation exists but should be rare, and happens when your fuel related executable exists, is in some other path and not in `.cargo/bin` (eg. `.local/bin`). In this case the warning only shows up and does not suggest any action to be taken.

Example:

image


bingcicle:
Added a pretty comprehensive test for the warning messages which may come in 4 different scenarios.

src/path.rs
```diff
@@ -51,3 +54,81 @@ pub fn ensure_dir_exists(path: &Path) -> Result<()> {
...
+                let mut message = String::new();
+                let fuelup_bin_dir = fuelup_bin_dir();
+                if !path.contains(fuelup_bin_dir.to_str().unwrap()) {
+                    let maybe_fuelup_executable = fuelup_bin_dir.join(&e);
+
+                    message.push_str(&format!("warning: '{e}' found in PATH at {path}."));
+
+                    if is_executable(&maybe_fuelup_executable) {
+                        message.push_str(&format!(
+                            " '{}' already installed at {} which will be overshadowed by the copy at {}.",
```
mitchmindtree, line 87:
```suggestion
                            " This will take precedence over '{}', already installed at {}. Consider uninstalling {}, or re-arranging your PATH to give fuelup priority.",
```
Maybe this is a little more consistent with the other branch's warning message? It also adds a general suggestion for uninstalling the other version or re-arranging PATH. Maybe we could add this suggestion to the other branch too?

mitchmindtree (commented):
Nice one @bingcicle, I feel like this might single-handedly solve 50% of devrel bugs/reports (provided users actually read `fuelup`'s output) :smiling_face_with_tear:

PR #302, opened 2022-12-06
warning message for trying to add a default forc plugin

bingcicle:


component/src/lib.rs
```diff
@@ -215,4 +224,17 @@ targets = ["linux_amd64", "linux_arm64", "darwin_amd64", "darwin_arm64"]
     fn test_collect_plugin_executables() {
         assert!(Components::collect_plugin_executables().is_ok());
     }
+
+    #[test]
+    fn test_is_default_forc_plugin() {
+        let executables = Components::collect_plugin_executables().unwrap();
+
+        for executable in executables.iter() {
+            if ["forc-lsp", "forc-doc", "forc-client", "forc-fmt"].contains(&executable.as_str()) {
```
mitchmindtree, line 233: Do we not have consts for these?

component/src/lib.rs
```diff
@@ -48,6 +48,15 @@ impl Component {
...
             .and_then(|c| Ok(c.clone()))
     }
+
+    pub fn is_default_forc_plugin(name: &str) -> bool {
+        (Self::from_name(FORC)
+            .unwrap()
+            .executables
+            .contains(&name.to_string())
+            && name != FORC)
+            || name == "forc-client"
```
mitchmindtree, line 58: same here

component/src/lib.rs
```diff
@@ -48,6 +48,15 @@ impl Component {
             .ok_or_else(|| anyhow!("component with name '{}' does not exist", name))
             .and_then(|c| Ok(c.clone()))
     }
+
+    pub fn is_default_forc_plugin(name: &str) -> bool {
+        (Self::from_name(FORC)
+            .unwrap()
```
mitchmindtree, line 54:
```suggestion
            .expect("there must always be a `forc` component")
```

component/src/lib.rs
bingcicle: Hmm, on second thought I might just remove this test because it's basically just testing a bool, which isn't very valuable, but the consts were removed when we introduced the `Component` abstraction which basically got rid of all the const strs that we were using.

component/src/lib.rs
bingcicle: Hmm, I didn't want to create a const for just 1 check at a place - the line of thought was that this could cause future PRs/contributions to use this `FORC_CLIENT` const should we add it to `constants.rs`, when this is just a one time check, plus it's probably unlikely that the name `forc-client` would be changed frequently anyway :thinking:

mitchmindtree: Using `const`s for str literals like this is not only about ease of refactoring or renaming, but also about removing the scope for human error when inevitably we want to use the same string again somewhere else.

E.g. it can be easy to miss a typo in a string literal, but you cannot write a typo in a `const` binding without getting a compile error. By maintaining a policy of using `const`s for literals rather than repeating the literals themselves, you only have to check for typos in one location.

> I didn't want to create a const for just 1 check at a place

We use the same literal above in the `executables` iteration :smiling_face_with_tear:

mitchmindtree (approved):
Approving as the last suggestion isn't major, will leave it up to you!

component/src/lib.rs
bingcicle: You make a good point, will change it accordingly :+1:

kayagokalp (approved):

PR #307, opened 2022-12-07
ci: add job to update channel on release

### LINKED ISSUE #227: automate updating channels upon change to `components.toml`
A change to `components.toml` should trigger a change in the `latest` channel as well.

Example:

1. add or remove `forc-wallet` to `components.toml`, make a PR.
2. once a change is detected in components.toml, a PR is automatically opened, which re-runs the publish script to include `forc-wallet` within the channel and commits the updates, whether we add/delete some lines or not.

bingcicle:
closes #227 

This is another way of fixing the issue detailed above. We do not employ the methods described above because changes to the channel without a new release of `fuelup` doesn't change how the existing `fuelup` would work.

Instead, this adds an extra step on release to re-publish the latest channel only done on release.


mitchmindtree (approved):
utACK

PR #311, opened 2022-12-12
tweak: do not consume download Err

bingcicle:
realized while looking at logs from someone reporting errors on discord, it's hard to know for sure the error when it's ignored.


kayagokalp (approved):
utACK

PR #314, opened 2022-12-14
fix: prefer .last() over .nth(1) while parsing version strings

bingcicle:
This was unsound code since the result of `fuel-indexer --version` is separated by more than 1 whitespace. We haven't ran into this issue so far since the other bins all have 1 whitespace separation between its version string and its bin name.

### DIFF
--- a/src/ops/fuelup_check.rs
+++ b/src/ops/fuelup_check.rs
@@ -57,7 +57,7 @@ fn check_plugin(plugin_executable: &Path, plugin: &str, latest_version: &Version
     {
         Ok(o) => {
             let output = String::from_utf8_lossy(&o.stdout).into_owned();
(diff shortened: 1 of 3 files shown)


mitchmindtree (approved):
The change looks fine, but I'm not sure I'm following the reasoning? Note that `split_whitespace` will skip all consecutive whitespace until the next non-whitespace character (see the second example here https://doc.rust-lang.org/std/primitive.str.html#method.split_whitespace).

bingcicle:
@mitchmindtree Here's the output for `fuel-indexer --version`:

```console
$ fuel-indexer --version
Indexer Service 0.1.10
```

Using nth(1) means that fuelup was trying to parse "Service" as a `semver::Version` instead here.

PR #317, opened 2022-12-14
feat: allow usage of fuel-toolchain.toml in projects

### LINKED ISSUE #65: Support project specific toolchains with `fuel-toolchain.toml`
Following some conversations on changes in forc from 0.14.5 -> 0.15.1 causing SwaySwap to break, I think there's space for fuelup to support usage of project-specific toolchains (which is basically rustup's override mechanic, described [here](https://rust-lang.github.io/rustup/overrides.html)). 

Will expand on this when I have more thoughts regarding this but my preliminary idea is that, instead of having a `rust-toolchain.toml`, we can have an entry within `Forc.toml` (would this create issues with parsing the toml?) stating which version of forc is built with the project. `fuelup` can detect this and decide which version of `forc` to run for that project. 

I think this idea is also mentioned [here](https://github.com/FuelLabs/fuelup/issues/1#issue-1242131332) in the original issue created by @adlerjohn. 

We would need to support the toolchain pattern #63 for this.

cc: @mitchmindtree @Braqzen for comments and ideas

bingcicle:
closes #65 

## Summary

This should be good to go for review. Basically this PR implements
- the store,
- basic overriding with `channel` and `components`.
- displaying the overridden toolchain with `fuelup default`.

### How does this work?
Whenever we execute binaries installed via fuelup, we always check in `proxy_cli.rs` the existence of an override file in the project root. From here, `fuelup` will decide which toolchain or component to use, and decide to install components (if not installed).

To test this PR:
1. checkout this PR, `cargo build`
2. link the built fuelup and a chosen fuelup binary of your choice to test. For me I am using `forc` and have this script to quickly iterate/test things:

```sh
rm -f ~/.fuelup/bin/forc
ln target/debug/fuelup ~/.fuelup/bin/forc
```

3. create a `fuel-toolchain.toml` at your current working directory, edit it to your wish to test things. You can see the [tests](https://github.com/FuelLabs/fuelup/blob/10e2807e5ef2b7470932a35e109002e2009da87d/src/toolchain_override.rs#L144) for a few examples.

This PR does **NOT** yet support:
- specifying the dates (eg. `nightly-2022-01-09` won't work, and by extension the dated latest versions) https://github.com/FuelLabs/fuelup/issues/331
- installing to store via `fuelup component add`. Ultimately the goal is to move all downloads to the store and link components from there. https://github.com/FuelLabs/fuelup/issues/329
- `fuelup show` - currently it behaves like before, but we want to show if something is overriding. https://github.com/FuelLabs/fuelup/issues/330

The reason the above few features aren't supported in this PR is for reviewability. Keeping the diff small would make reviews easier. The above 3 features will be implemented in the next few PRs.

~Opening a draft PR to track the deliverables below:~

## Todos

- [x] fuelup should adhere to both the naming conventions of distributed toolchains and custom toolchains
- [x] fuelup should auto download the toolchain, or any missing components, when any `fuelup` managed binaries are invoked with this toml file in the project root
- [x] ~`fuelup show` should show the override as the currently active toolchain, instead of what is shown in `fuelup default`~ deferred to separate PR for reviewability
- [x] `fuelup default` should show current override as well
- [x] document `fuel-toolchain.toml` in the fuelup book
- [x] add some tests with `fuel-toolchain.toml`

Additional todos, following updates [here](https://github.com/FuelLabs/fuelup/issues/320):
- [x] download missing components to a new directory, `/.fuelup/store`, instead of in the toolchain context
- [x] execute components from the store directly instead

### DIFF
--- a/docs/src/overrides.md
+++ b/docs/src/overrides.md
@@ -10,7 +10,7 @@ You can override the installed default toolchain using a `fuel-toolchain.toml` f
 Using the `fuel-toolchain.toml` file is a way to have projects 'pinned' to specific versions of components
 in the Fuel toolchain and have this information reflected in their source repository.
 
-When this file is present, `fuelup` knows to override the default toolchain with the specified toolchain when executing binaries
+When this file is present, `fuelup` will override the default toolchain with the specified toolchain when executing binaries
 in the toolchain.
 
 In these cases, the toolchain can be specified in a file called `fuel-toolchain.toml`. These toolchains can only be
--- a/docs/src/overrides.md
+++ b/docs/src/overrides.md
@@ -5,7 +5,7 @@
 
 Currently, the only way of overriding toolchains is through the [`fuel-toolchain.toml`] file.
 
-# The toolchain file
+## The toolchain file
 
 Using the `fuel-toolchain.toml` file is a way to have projects 'pinned' to specific versions of Sway and have this information reflected in their source repository.
 
--- a/docs/src/overrides.md
+++ b/docs/src/overrides.md
@@ -38,7 +38,7 @@ An application using the [`beta-2`] toolchain:
 channel = "beta-2"
 ```
 
-Let's say we have a project on the `beta-2` testnet, and we want to try using the 
+Let's say we have a project on the `beta-2` testnet, and we want to try using the
 latest (at time of writing) forc to develop on it:
 
 ```toml
--- a/docs/src/overrides.md
+++ b/docs/src/overrides.md
@@ -3,7 +3,7 @@
 `fuelup` automatically determines which [toolchain] to use when one of the installed commands like
 `forc` is executed.
 
-Currently, the only way of overriding toolchains is through the `fuel-toolchain.toml` file.
+You can override the installed default toolchain using a `fuel-toolchain.toml` file.
 
 ## The toolchain file
 
--- a/docs/src/overrides.md
+++ b/docs/src/overrides.md
@@ -58,4 +58,3 @@ components = []
 [`fuel-toolchain.toml`]: overrides.md#the-toolchain-file
 [custom toolchain]: concepts/toolchains.md#custom-toolchains
 [`latest`]: concepts/channels/latest.md
-
--- a/docs/src/overrides.md
+++ b/docs/src/overrides.md
@@ -3,17 +3,18 @@
 `fuelup` automatically determines which [toolchain] to use when one of the installed commands like
 `forc` is executed.
 
-Currently, the only way of overriding toolchains is through the [`fuel-toolchain.toml`] file.
+Currently, the only way of overriding toolchains is through the `fuel-toolchain.toml` file.
 
 ## The toolchain file
 
-Using the `fuel-toolchain.toml` file is a way to have projects 'pinned' to specific versions of Sway and have this information reflected in their source repository.
+Using the `fuel-toolchain.toml` file is a way to have projects 'pinned' to specific versions of components 
+in the Fuel toolchain and have this information reflected in their source repository.
 
 When this file is present, `fuelup` knows to override the default toolchain with the specified toolchain when executing binaries
 in the toolchain.
 
-In these cases, the toolchain can be specified in a file called `fuel-toolchain.toml`. These toolchains can either be
-the [distributed toolchains or custom toolchains].
+In these cases, the toolchain can be specified in a file called `fuel-toolchain.toml`. These toolchains can only be
+the [distributed toolchains] at this point in time.
 
 Here's what a sample project might look like:
 
@@ -28,33 +29,25 @@ $ tree -L 1 # 'tree -L 1' shows the project structure up to the depth of 1
 └── SPECIFICATION.md
 ```
 
-And the `fuel-toolchain.toml` for an AMM application using a [custom toolchain] might contain:
+### Examples
+
+1. an application using the [`beta-2`] toolchain:
 
 ```toml
 [toolchain]
-name = "my-awesome-amm-toolchain"
-components = ["forc", "fuel-core", "forc-index", "fuel-indexer"]
+channel = "beta-2"
 ```
 
-Note that the components key itself and the array value are both optional.
-
-For example, an application using the [`latest`] toolchain might contain:
+2. An application using the [`beta-2`] toolchain, but using another version of forc:
 
 ```toml
 [toolchain]
-name = "latest"
-```
+channel = "beta-2"
 
-or:
-
-```toml
-[toolchain]
-name = "latest"
-components = []
+[components]
+forc = "0.32.2" # in beta-2, forc is pinned to v0.31.1
 ```
 
 [toolchain]: concepts/toolchains.md
-[distributed toolchains or custom toolchains]: concepts/toolchains.md#toolchains
-[`fuel-toolchain.toml`]: overrides.md#the-toolchain-file
-[custom toolchain]: concepts/toolchains.md#custom-toolchains
-[`latest`]: concepts/channels/latest.md
+[distributed toolchains]: concepts/toolchains.md#toolchains
+[`beta-2`]: concepts/channels/beta-2.md
--- a/docs/src/overrides.md
+++ b/docs/src/overrides.md
@@ -7,8 +7,9 @@ You can override the installed default toolchain using a `fuel-toolchain.toml` f
 
 ## The toolchain file
 
-Using the `fuel-toolchain.toml` file is a way to have projects 'pinned' to specific versions of components
-in the Fuel toolchain and have this information reflected in their source repository.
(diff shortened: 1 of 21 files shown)


bingcicle:
This should be gtg for review, but note that the `components` key do not allow for specifying for specific versions in this PR - this will be handled in a next PR. Issue [here](https://github.com/FuelLabs/fuelup/issues/320)

src/toolchain_override.rs
```diff
@@ -0,0 +1,82 @@
...
+            ),
+            Some(components) => {
+                for component in components {
+                    if !toolchain.has_component(component) {
+                        let target_triple = TargetTriple::from_component(component).unwrap_or_else(|_| {
+                            panic!("Failed to create target triple for '{}'", component)
+                        });
+
+                        if let Ok(download_cfg) = DownloadCfg::new(called, target_triple, None) {
+                            toolchain.add_component(download_cfg).unwrap_or_else(|_| {
```
mitchmindtree, line 69: Perhaps we should provide an `info!` to let users know we were missing a component and that fuelup is downloading it? E.g. otherwise a user might run something like `forc build` or `fuel-core` and it might take a strangely long amount of time without any explanation.

src/toolchain_override.rs
```diff
@@ -0,0 +1,82 @@
...
+        match self.toolchain.components.as_deref() {
+            Some([]) | None => warn!(
+                "warning: overriding toolchain '{}' in fuel-toolchain.toml does not have any components listed",
+                &self.toolchain.name
+            ),
+            Some(components) => {
+                for component in components {
+                    if !toolchain.has_component(component) {
+                        let target_triple = TargetTriple::from_component(component).unwrap_or_else(|_| {
+                            panic!("Failed to create target triple for '{}'", component)
```
mitchmindtree, line 65: Can we return an error rather than panic here?

src/toolchain_override.rs
```diff
@@ -0,0 +1,82 @@
...
+                        });
+
+                        if let Ok(download_cfg) = DownloadCfg::new(called, target_triple, None) {
+                            toolchain.add_component(download_cfg).unwrap_or_else(|_| {
+                                panic!(
+                                    "Failed to add component '{}' to toolchain '{}'",
+                                    component, toolchain.name,
+                                )
```
mitchmindtree, line 73: Same here?

src/toolchain_override.rs
bingcicle: Yep, thanks for this catch. The intent was originally to disallow execution of commands if there existed a corrupted `fuel-toolchain.toml` in the project root, hence the `panic!`. But this is probably bad design anyway on hindsight.

src/toolchain_override.rs
bingcicle: The `add_component` fn already has built in `info!` spans telling the user that fuelup is downloading a component (hurray for reusable code), but I'll still add an additional info span here to note that we are installing from a missing component that was specified in `fuel-toolchain.toml`. This case was originally meant for supporting custom toolchains, since `component add/remove` are disallowed in a distributed toolchain anyway.

docs/src/overrides.md
```diff
@@ -0,0 +1,60 @@
...
+├── README.md
+└── SPECIFICATION.md
+```
+
+And the `fuel-toolchain.toml` for an AMM application using a [custom toolchain] might contain:
+
+```toml
+[toolchain]
+name = "my-awesome-amm-toolchain"
+components = ["forc", "fuel-core", "forc-index", "fuel-indexer"]
```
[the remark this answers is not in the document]
bingcicle: Following a discussion with @mitchmindtree the decision is to follow this format: ([link to comment](https://github.com/FuelLabs/fuelup/issues/65#issuecomment-1358923366))

> Updated proposed look of `fuel-toolchain.toml`:
> 
> ```toml
> [toolchain]
> channel = "latest"
> ```
> 
> This pins the current project toolchain to `latest`.
> 
> the TOML also accepts an optional table `components`, if you wish to override certain components from the toolchain:
> 
> ```toml
> [toolchain]
> channel = "latest"
> 
> [components]
> forc = "0.31.3"
> ```
> 
> This means that all other components will be derived from the `latest` toolchain.

bingcicle:
changing back to a draft PR while I fix the format of the TOML file.

bingcicle:
OK, I think this should be  good to go for review - this was tricky to test and reason about because of the conflicting behaviours of how fuelup currently reasons about toolchains vs the new store model. Hopefully I caught most of the cases here.

src/proxy_cli.rs
```diff
@@ -25,22 +28,61 @@ pub fn proxy_run(arg0: &str) -> Result<ExitCode> {
...
+    let (bin_path, toolchain_name) = match toolchain_override {
+        Some(to) => {
+            // unwrap() is safe here since we try DistToolchainDescription::from_str()
+            // when deserializing from the toml.
+            let description =
+                DistToolchainDescription::from_str(&to.cfg.toolchain.channel).unwrap();
+
+            // Since we have a valid
+            let toolchain = Toolchain::from_path(&description.to_string()).unwrap_or_else(|_| {
+                panic!("Failed to create toolchain '{}' from path", &description)
```
bingcicle, line 42: Note: I'm using `panic!` here and below for 2 reasons:

1) this is the point of entry into fuelup-managed commands like `forc`, so we shouldn't continue with execution if we end up with bad state here, and
2) for this particular fn, creating toolchain from path that we've already done some validation above shouldn't fail.

mitchmindtree: > this is the point of entry into fuelup-managed commands like forc, so we shouldn't continue with execution if we end up with bad state here

This doesn't imply we should `panic!`, but rather that we should make sure when this error is raised that we actually exist the program (and don't try to continue).

> for this particular fn, creating toolchain from path that we've already done some validation above shouldn't fail.

This on the other hand is a good reasons to `panic!`, i.e. if we're checking something that we *know* has already been checked.

That said, it looks like this `Toolchain` constructor does `I/O` (reads from a file), so it doesn't really matter if we've already checked it before - it's always possible that since the last time we checked it, it's been deleted or corrupted by some other process or anything really, so technically we should probably handle this.

This is not a huge deal for fuelup, but just wanted to clarify good practise around producing errors vs `panic!`ing.

mitchmindtree (commented):
Didn't quite get to finish reviewing this yet, but thought I'd leave a few comments in the meantime! Looking forward to checking out the new store stuff :)

src/proxy_cli.rs
bingcicle: Actually I think there's a simpler solution - there wasn't any reason to return a `Result<Toolchain>` anyway since there wasn't anything upstream that returns a `Result`. Will create another PR to update the fn signature to return `Self` instead of `Result<Self>` (so we can keep things clean here)

I updated the `direct_proxy` fn signature to return `anyhow::Error` over `io::Error` (downstream `exec()` fn still returns io::Result), this probably makes things easier.

But definitely good points regarding `panic!` vs errors, will keep in mind :+1:

src/store.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    pub(crate) fn has_component(&self, component_name: &str, version: &Version) -> bool {
+        let dirname = self.component_dirname(component_name, version);
+        self.path().join(dirname).exists()
+    }
+
+    pub(crate) fn component_dirname(&self, component_name: &str, version: &Version) -> String {
+        format!("{component_name}-{version}")
+    }
```
mitchmindtree, line 35: It looks like this could be a free function as it doesn't require `self`.

src/toolchain_override.rs
```diff
@@ -0,0 +1,218 @@
...
+        if let Some(components) = self.components {
+            document["components"] = toml_edit::Item::Table(toml_edit::Table::new());
+            for (k, v) in components.iter() {
+                document["components"][k] = value(v.to_string());
+            }
+        }
+
+        document
+    }
+}
```
mitchmindtree, line 150: Rather than having manual `from_toml` and `to_document` implementations, is it possible to have the derived `Deserialize` or `Serialize` do the conversions as we'd expect?

mitchmindtree (commented):
Nice work, looks great! Just a couple last comments.

Looking forward to seeing #329 addressed with a more thorough use of the store throughout fuelup :)

src/toolchain_override.rs
bingcicle: The derived `Deserialize`/`Serialize` outputs them not in the format that we want. 

I dug a little into the `toml_edit` docs and seems like `to_string_pretty()` ([link](https://docs.rs/toml_edit/latest/toml_edit/ser/fn.to_string_pretty.html)) supports the use case that I want for tests here.

mitchmindtree (approved):
:ship:

mitchmindtree (approved):

PR #318, opened 2022-12-16
feat: apps repo toolchain

bingcicle:
This is mainly for the CDMX hackathon happening Dec 15-16  to allow hackers to run examples from https://github.com/FuelLabs/sway-applications easily.


bingcicle:
Closing this instead so we don't maintain too many 'official' toolchains. This was a one-time thing and upon further investigation the versions that the app-repo was using was the same as `beta-2`, so the suggestion was to use that instead.

Issue #321, opened 2022-12-19
feat: lock components used in `fuel-toolchain.toml`


bingcicle:
Following #320 there needs to be a way for projects to lock and update the toolchain.

The current mechanism for detecting updates on distributed toolchains is by checking the hash of the channel in `~/.fuelup/hashes/<toolchain>`. When a channel on GitHub is updated, this hash changes and `fuelup` knows to update if the hash doesn't match. We can use the same mechanism for `fuel-toolchain.toml` - though this might introduce some complexity:

- clash with `fuelup component add`: the add command should still add to the default toolchain instead of the override.
- updating the override will have to be through editing `fuel-toolchain.toml`.

This is mainly due to convenience - editing the TOML file will result in the hash changing which makes it easier for devs to make local changes to the TOML file and update components.

mitchmindtree:
We may want to consider locking using a separate file (e.g. `fuel-toolchain.lock`). Then we can use a flow similar to forc's, e.g.

1. If `fuel-toolchain.toml` exists, but `fuel-toolchain.lock` does not yet exist, select the latest versions based compatible with the `fuel-toolchain.toml`. Write the lock file with these versions.
2. If both the `fuel-toolchain.toml` and `fuel-toolchain.lock` exist, check for any incompatibilities in components or versions between the toml and lock (can occur if the toml was edited since the lock was generated). If so, determine new versions and update the lock.
3. Use the exact versions specified in the resulting lock.

Each time we update the lock, we should make it clear to the user that we do so (like we do in forc).

We could provide a `fuelup lock update` command or something like this to make it easier to update to the latest patch releases?

bingcicle:
This is closed by #335 since we now publish `latest` with dates, and #317 which introduces the store in `~/.fuelup/store`, which dictates that `fuelup` now installs and uses components from the store. A lockfile would be unnecessary in this paradigm - assuming that a team is working on the same project and one member pushes changes to `fuel-toolchain.toml`, another member who pulls the changes would be able to grab updates. This means that the toml itself acts as the lockfile.

PR #323, opened 2022-12-19
feat: fuel toolchain specify version

### LINKED ISSUE #320: feat: allow specifying distributed toolchains with overriding versions in `fuel-toolchain.toml`
Follow-up to https://github.com/FuelLabs/fuelup/pull/317

Currently in the `fuel-toolchain.toml` file, you can only specify components without specific versions and they download the `latest` versions:

```toml
[toolchain]
name = "my-toolchain"
components = ["forc"]
```

We want to allow:

```toml
[toolchain]
name = "my-toolchain"
components = ["forc@0.31.3"]
```

bingcicle:
closes #320 

This might be far more complicated than it needed to be - ended up writing a serializer/deserializer for a `Component` struct, which contains a `name: String` and a `version: Option<semver::Version>`.

@mitchmindtree 's suggested TOML format [here](https://github.com/FuelLabs/fuelup/issues/320#issuecomment-1357169806) might reduce some complexity in parsing the file but make it slightly more verbose for the DevEx. I think this approach is a slightly nicer way to keep everything hidden from sight, but am open to changing the design/implementation at this stage before this is set in stone.

We also have to keep in mind #321 for whichever approach ends up being the final one, as this PR will affect how #321 is handled.

### DIFF
--- a/src/ops/fuelup_component/add.rs
+++ b/src/ops/fuelup_component/add.rs
@@ -10,6 +10,25 @@ use crate::{
     toolchain::Toolchain,
 };
 
+pub fn split_versioned_component(
+    maybe_versioned_component: &str,
+) -> Result<(String, Option<Version>)> {
+    match maybe_versioned_component.split_once('@') {
+        Some(t) => {
+            let v = match Version::from_str(t.1) {
+                Ok(v) => Some(v),
+                Err(e) => bail!(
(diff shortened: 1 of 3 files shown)


bingcicle:
Closing this PR in favor of the new format introduced in #317

PR #328, opened 2023-01-04
ci: publish latest twice to include one with date

### LINKED ISSUE #324: CI: publish latest channels with dates
- Each channel should have a new key, `date`, where we publish when the channel was published.
- Whenever a new channel is published, we save the previous one in a subdirectory. To facilitate this, we should create a new subdirectory `/channels/latest`, and save the older versions there.

For example, in our current setup, lets say a new channel is published today eg. 2022-12-22 (YYYYMMDD).  ~The CI script should read the `date` key of the previous channel, move the previous channel into `/channels/latest` and append the date in the file name, eg. `channel-fuel-latest-2022-12-22.toml`.~

EDIT: Instead of the above complicated way of doing this we should just publish once in 2 locations whenever there is a new `latest`, no need for checking for the previous versions.

The new channel will be published into two places - as the default `latest` channel (channel-fuel-latest) and in the date-appended format as well.

- In making the above changes, we should also probably move all channels in the base gh-pages branch into the already existing `/channels` directory.

bingcicle:
closes #324 

2 changes with this PR:

- Updates the toolchain compatibility workflow to publish the `latest` channel twice: once within the root of gh-pages as `channel-fuel-latest.toml` (the same format as we have now), and another time within `/channels/latest` as `channel-fuel-latest-YYYY-MM-DD.toml` (to be used for `fuel-toolchain.toml`).

- Previously used $GITHUB_RUN_ID as an input into the `build-channel` script, now we're adding the date as well and there are too many args into the script for my liking. Using `sed` to prepend the `published_by` and `date` metadata into the channel instead.

### DIFF
--- a/.github/workflows/publish-nightly-channel.yml
+++ b/.github/workflows/publish-nightly-channel.yml
@@ -33,7 +33,7 @@ jobs:
 
             CHANNEL_TOML="channel-fuel-nightly.toml"
 
-            build-channel nightly $CHANNEL_TOML $GITHUB_RUN_ID
+            build-channel nightly $CHANNEL_TOML
 
             cp $CHANNEL_TOML ${{ env.NIGHTLY_CHANNEL_DIR }}
 
--- a/.github/workflows/publish-nightly-channel.yml
+++ b/.github/workflows/publish-nightly-channel.yml
@@ -28,16 +28,13 @@ jobs:
       - name: Publish nightly channel
         id: setup
         run: |
-            PUBLISHED_DATE=$(date + '%Y%m%d')
+            PUBLISHED_DATE=$(date + '%Y-%m-%d')
             FORMATTED_PUBLISHED_DATE=$(date +'%Y/%m/%d')
             mkdir -p ${{ env.NIGHTLY_CHANNEL_DIR }}
 
             CHANNEL_TOML="channel-fuel-nightly.toml"
 
-            build-channel nightly $CHANNEL_TOML
-
-            sed -i "1 i\published_by = \"https://github.com/FuelLabs/fuelup/actions/runs/${GITHUB_RUN_ID}\"\n" $CHANNEL_TOML
-            sed -i "1 i\date = \"${PUBLISHED_DATE}\"" $CHANNEL_TOML
+            build-channel nightly $CHANNEL_TOML $GITHUB_RUN_ID $PUBLISHED_DATE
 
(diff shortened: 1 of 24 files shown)


mitchmindtree (commented):
> Updates the toolchain compatibility workflow to publish the latest channel twice

I wonder if we should just publish the one with the date, and use that from the commands that would previously just check for the channel? Or is the issue that we don't know the latest date from within fuelup? Maybe we can redirect a request from the old location to the one with the latest date? Happy to land this approach for now if it's easier :+1:  

> Previously used $GITHUB_RUN_ID as an input into the build-channel script, now we're adding the date as well and there are too many args into the script for my liking. Using sed to prepend the published_by and date metadata into the channel instead.

I think we're better off picking one approach - either we write the toml in bash, or write it in Rust, but not an arbitrary mix of the two. To future readers/maintainers of this code, without finding this github comment, it wouldn't be clear why the `date` and `published_by` fields are edited using sed, when the rest is constructed using the Rust code. I'd probably opt for adding the extra args to the Rust CLI.

bingcicle:
> I wonder if we should just publish the one with the date, and use that from the commands that would previously just check for the channel? Or is the issue that we don't know the latest date from within fuelup? Maybe we can redirect a request from the old location to the one with the latest date? 

Yeah `fuelup`'s awareness of `latest` is only just that, it doesn't know the date prior to the changes here. Also, `channel-fuel-latest.toml` is a const hardcoded into fuelup, so if we abandoned the old format and deleted the toml from our gh-pages, this would break older versions of fuelup (though, i supposed there's no reason not to have fuelup on the latest version)

> it wouldn't be clear why the date and published_by fields are edited using sed, when the rest is constructed using the Rust code

This makes sense, will update the behaviour

mitchmindtree (approved):
Noice :shipit:

bingcicle:
To integrate better with #335 and for consistency reasons with `nightly`, I'm changing the publish location of the dated `latest` tomls to be the same as nightly:

```console
$ tree channels/ -L 3
channels/
└── nightly
    └── 2022
        ├── 09 
        ├── 10
        └── 11
```

The new publish location, if for example there was a new `latest` today:

```console
$ tree channels/ -L 3
channels/
...
└── latest
    └── 2023 # year
        ├── 01 # month
        │   ├── 11 # day
                ├── channel-fuel-latest-2023-01-11.toml
```

mitchmindtree (approved):

PR #332, opened 2023-01-09
feat: `fuelup show` should show the overridden toolchain

### LINKED ISSUE #330: feat: support fuelup show when override is active
We should show the override toolchain, its active components, and the path of the `fuel-toolchain.toml` file when an override is active.

bingcicle:
Closes #330 

Example, with `fuel-toolchain.toml` declaring `beta-2` as the overriding toolchain:

image


mitchmindtree (approved):
utACK

PR #334, opened 2023-01-10
refactor: update `Toolchain::from_path` to return `Self` instead of `Result<Self>`

bingcicle:
Discovered here: https://github.com/FuelLabs/fuelup/pull/317#discussion_r1065218363


mitchmindtree (approved):
Niccce

PR #335, opened 2023-01-10
feat(fuel-toolchain.toml): support dates

### LINKED ISSUE #331: feat: support specifying dates for `latest` and `nightly`
Currently `fuel-toolchain.toml` does not support specifying dates as its `channel`.

bingcicle:
closes #331

~Marked this as ready prematurely but converting back to a draft for now while I work on possibly refactoring parts of the `Toolchain` and `Channel` abstractions. There are various overlapping concepts here as a result of mostly forking the functionality of `rustup`, going to leave my thoughts below:~

~We currently have `DistToolchainDescription` to reason about toolchains - it usually comes in the form of `<name>-<date>-<target>`, with `date` and `target` being optional. It is used as a way to parse user input. It is valid for users to do `fuelup toolchain install latest`, and also `fuelup toolchain install nightly-2023-01-10`.~

~Currently `latest` has no knowledge or concept of dates - so this has to be updated. This includes various places of the codebase.~

EDIT:

This should be ready for review.

## Overview

We're using `DistToolchainDescription` whereever possible to display and parse toolchain inputs; because of this abstraction thankfully we're easily able to enforce dated `latest`s with this PR. The concepts surrounding `DistToolchainDescription` do seem a bit messy atm and I think this should call for some minor refactoring in future, but I'll leave that for a later PR.

## How to test?

First put this convenience script `build.sh` in fuelup dir:

```sh
cargo build
rm -f ~/.fuelup/bin/forc
rm -f ~/.fuelup/bin/fuelup
ln target/debug/fuelup ~/.fuelup/bin/forc
ln target/debug/fuelup ~/.fuelup/bin/fuelup
```

Use this to run cargo build and it should automatically link dev fuelup that handles the new logic from this PR. Note that this messes up your actual `forc` and `fuelup` so you would have to reinstall fuelup after you're done reviewing.

Create a `fuel-toolchain.toml`, declaring `latest-2023-01-06`:

```toml
[toolchain]
channel = "latest-2023-01-06"
```

This channel was manually deployed by me (we don't have any new versions yet, but I used the date of the last release which was forc v0.33)

Assuming you've ran the above `build.sh`, now try running `forc`. If you don't have the above toolchain installed, `fuelup` will now fetch and install it for you.

Now try it again with a random dated nightly:

```toml
[toolchain]
channel = "nightly-2022-09-23"
```

Now try using `latest` and `nightly` (without dates), this should warn you that the toml is invalid, and default to using your default toolchain's `forc`.

### DIFF
--- a/src/toolchain_override.rs
+++ b/src/toolchain_override.rs
@@ -18,8 +18,6 @@ use crate::{
     path::get_fuel_toolchain_toml, target_triple::TargetTriple, toolchain::Toolchain,
 };
 
-const CHANNEL_WITHOUT_DATE_ERROR: &str = "channel with date";
-
 // For composability with other functionality of fuelup, we want to add
 // additional info to OverrideCfg (representation of 'fuel-toolchain.toml').
 // In this case, we want the path to the toml file. More info might be
@@ -87,7 +85,7 @@ where
     if channel_str == LATEST || channel_str == NIGHTLY {
         return Err(Error::invalid_value(
             serde::de::Unexpected::Str(&channel_str),
-            &CHANNEL_WITHOUT_DATE_ERROR,
+            &"channel with date",
         ));
     }
 
--- a/src/toolchain_override.rs
+++ b/src/toolchain_override.rs
@@ -15,65 +15,32 @@ use crate::{
     toolchain::{DistToolchainName, Toolchain},
 };
 
+// For composability with other functionality of fuelup, we want to add
+// additional info to OverrideCfg (representation of 'fuel-toolchain.toml').
+// In this case, we want the path to the toml file. More info might be
+// needed in future.
 #[derive(Debug, Deserialize, Serialize)]
 pub struct ToolchainOverride {
     pub cfg: OverrideCfg,
     pub path: PathBuf,
 }
 
+// Representation of the entire 'fuel-toolchain.toml'.
 #[derive(Debug, Deserialize, Serialize)]
 pub struct OverrideCfg {
     pub toolchain: ToolchainCfg,
     pub components: Option<HashMap<String, Version>>,
 }
 
+// Represents the [toolchain] table in 'fuel-toolchain.toml'.
 #[derive(Debug, Deserialize, Serialize)]
 pub struct ToolchainCfg {
     pub channel: String,
 }
 
-impl OverrideCfg {
-    pub fn new(toolchain: ToolchainCfg, components: Option<HashMap<String, Version>>) -> Self {
-        Self {
-            toolchain,
-            components,
-        }
-    }
-
-    pub fn to_document(self) -> Document {
-        let mut document = toml_edit::Document::new();
-
-        document["toolchain"] = toml_edit::Item::Table(toml_edit::Table::new());
-        document["toolchain"]["channel"] = value(self.toolchain.channel);
-        if let Some(components) = self.components {
-            document["components"] = toml_edit::Item::Table(toml_edit::Table::new());
-            for (k, v) in components.iter() {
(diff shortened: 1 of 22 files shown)


src/toolchain_override.rs
```diff
@@ -1,18 +1,21 @@
...
+use crate::constants::DATE_FORMAT;
+use crate::toolchain::DistToolchainDescription;
 use crate::{
-    constants::FUEL_TOOLCHAIN_TOML_FILE,
-    download::DownloadCfg,
-    file,
-    path::get_fuel_toolchain_toml,
-    target_triple::TargetTriple,
-    toolchain::{DistToolchainName, Toolchain},
+    constants::FUEL_TOOLCHAIN_TOML_FILE, download::DownloadCfg, file,
```
mitchmindtree, line 17: Above we have `use crate::constants::DATE_FORMAT`, but then we also import from `constants` in this import line? We should pick one approach and try to be more consistent with the way we import items. Not a big deal - maybe something `cargo fmt` can do for us automatically in the future.

src/toolchain_override.rs
```diff
@@ -33,9 +36,96 @@ pub struct OverrideCfg {
 }
 
 // Represents the [toolchain] table in 'fuel-toolchain.toml'.
-#[derive(Debug, Deserialize, Serialize)]
+#[derive(Debug, Deserialize)]
 pub struct ToolchainCfg {
-    pub channel: String,
+    #[serde(deserialize_with = "deserialize_channel")]
```
mitchmindtree, line 41: An alternative to providing a function like this is to implement `Deserialize` for `Channel` manually. This can be useful if `Channel` is used in some other serializable type and we want it to have the same serialization behaviour there too. Not important in this case, just a tip!

src/toolchain_override.rs
```diff
@@ -33,9 +36,96 @@ pub struct OverrideCfg {
...
+                date: Date::parse(d, DATE_FORMAT).ok(),
+            })
+        } else {
+            if s == LATEST || s == NIGHTLY {
+                bail!("'{s}' without date specifier is forbidden");
+            }
+            bail!("Invalid str for channel: '{}'", s);
+        }
+    }
 }
```
mitchmindtree, line 129: Can you re-use this `FromStr` implementation in the `deserialize_channel` function above? I.e. call `Channel::parse` within `deserialize_channel`? It looks like there's quite a bit of duplicated logic between the two.

mitchmindtree (dismissed):
Looks good! utACK. I've left some optional comments, happy to re-approve if you want to address them but also don't mind if they're left for future refactor.

src/toolchain_override.rs
bingcicle: Thanks for the catch! Cursed auto-complete doesn't seem to be able to reconcile nested imports with unnested ones.

Does `cargo fmt` catch this? Didn't seem to work for me.

src/toolchain_override.rs
bingcicle: I saw that there seemed to be more boilerplate code when implementing `Deserialize`, so I opted to do this instead. I'll create an issue for this, it's probably useful for the future as well

src/toolchain_override.rs
mitchmindtree: Ahh I meant that hopefully in the future cargo fmt would be able to catch this for us - I'm not sure if it can now, but I don't think it does by default at least.

mitchmindtree (approved):
:shipit:

bingcicle:
sorry - had to fix a clippy lint + merge conflict :sweat:  @mitchmindtree

mitchmindtree (approved):

PR #337, opened 2023-01-11
refactor: update nightly publishes

### LINKED ISSUE #338: ci: nightly channel publishes failing silently
[Some](https://github.com/FuelLabs/fuelup/blob/gh-pages/channels/nightly/2023/01/10/channel-fuel-nightly.toml) of the latest [nightly](https://github.com/FuelLabs/fuelup/blob/gh-pages/channels/nightly/2023/01/11/channel-fuel-nightly.toml) publishes succeeded but failed silently - you can see that some versions were declared but there are no download links available for some components.

bingcicle:
Closes #338 

[Some](https://github.com/FuelLabs/fuelup/blob/gh-pages/channels/nightly/2023/01/10/channel-fuel-nightly.toml) of the latest [nightly](https://github.com/FuelLabs/fuelup/blob/gh-pages/channels/nightly/2023/01/11/channel-fuel-nightly.toml) publishes succeeded but failed silently - you can see that some versions were declared but there are no download links available for some components. Previously this CI workflow had to manually build all the strings for nightly releases - this led to the issues above, where it seems like git commit hashes did not match at the time of publish versus the time of release. Generally the way the script worked led to unreliability of the publish workflow because nightly builds can fail for a variety of reasons.

## Reviewing this PR

Because of https://github.com/FuelLabs/sway-nightly-binaries/pull/32 unifying all nightly releases into 1, we can build the `nightly` channel depending on what is within the release instead of arbitrarily building strings/URLs as described above. 2 main areas to note when reviewing this PR:

1) Most of the logic changes to review is in `write_nightly_document()`. To summarize what that fn is doing, we're simply querying the github release API for the tag (which is just `nightly-<DATE>`), then iterating over the assets, and building the channel based on the assets - this improves reliability of the publish step, since we're not trying to publish anything that doesn't actually exist.

2) For readability reasons of the script I split the big `main()` from before based on which channel the script is trying to publish for. `write_latest_document()` shouldn't have much changes logic-wise, only minor clean up things removing `nightly` publishing steps. I also went ahead and removed some unused structs.


ci/build-channel/src/main.rs
```diff
@@ -152,22 +118,68 @@ fn validate_components(channel: &str, components: &HashMap<String, Version>) ->
...
+
+                            println!(
+                                "url: {}\nhash: {}",
+                                &asset.browser_download_url, &actual_hash
+                            );
+                            document["pkg"][&component.name]["target"][target.to_string()]
+                                ["hash"] = value(actual_hash);
+                        };
+                    };
+                }
```
mitchmindtree, line 171: Rather than writing directly to the toml document, one alternative approach that might be a bit more type-safe is to create a Rust type to represent the data you want, populate that, then serialize that type to `toml` afterwards. This helps to avoid typos in strings, and you get the side-benefit of being able to use the same type for deserialization.

mitchmindtree (approved):
Looks good! I've left one suggestion but will leave it up to you :+1:

ci/build-channel/src/main.rs
bingcicle: I'll create an issue for this for now, but this might need changing up the flow/logic because of the organization of the toml file.

PR #340, opened 2023-01-13
fix: default bug

### LINKED ISSUE #339: fuelup: bug in the fuelup default beta-2 command
### Problem
  
  On running the fuelup default beta-2 command, it gives the following output:
  **No default toolchain detected. Please install or create a toolchain first.**

### Expected Behaviour

  The above command should ideally result in successfully setting up the beta-2 toolchain as the default toolchain with the following output:
**default toolchain set to 'beta-2-aarch64-apple-darwin'**

### Steps

1. Install fuelup on mac m2 for aarch64-apple-darwin
2. Run the **fuelup install toolchain beta-2** in the terminal
3. Run the **fuelup default beta-2** command

### Possible Solution(s)

_No response_

### Notes

The output of `fuelup show` is the following:

  

> Default host: aarch64-apple-darwin
>   fuelup home: /Users/vish/.fuelup
>   
>   installed toolchains
>   No default toolchain detected. Please install or create a toolchain first.

However, `ls ~/.fuelup/toolchains` gives the following output, which means the toolchain is installed:
  

> beta-2-aarch64-apple-darwin	latest-aarch64-apple-darwin

### Fuelup version

```console
fuelup 0.15.1
```

### Installed components

```console
latest-aarch64-apple-darwin
  forc - Up to date : 0.33.0
  fuel-core - Up to date : 0.15.1
  fuel-indexer - Up to date : 0.1.12
beta-2-aarch64-apple-darwin
  forc - Error getting version string
  fuel-core - Up to date : 0.14.1
fuelup - Up to date : 0.15.1
```

bingcicle:
closes #339 

So the cause of the bug was actually running `fuelup default` prior to installing any toolchain. As a result of the ordering of the code, a `settings.toml` file is generated but with no content inside, and on the next run of `fuelup default`, we always error out of the command, since `Toolchain::from_settings()` is the first fn call.

Changes:
- in `Toolchain::from_settings()` - check if `settings.toml` exists first.
- Reorder some code in `fuelup default`
- Added a line to assert that `settings.toml` isn't mistakenly created when `fuelup default` is ran on an empty setup.

There are 2 modes that `fuelup default` runs in. You can either execute:
1) `fuelup default`, or
2) `fuelup default <toolchain>`

The former prints out your current default toolchain and the latter switches your default toolchain. In the case of 2) we actually don't care if there's a current default toolchain - we only want to switch toolchains. Hence the move of this:

```rust
let current_toolchain = Toolchain::from_settings()?;
```
into the `Some` match statement, for when there is something to switch to.

### DIFF
--- a/src/ops/fuelup_default.rs
+++ b/src/ops/fuelup_default.rs
@@ -10,20 +10,24 @@ use crate::{
 };
 
 pub fn default(toolchain: Option<String>) -> Result<()> {
-    let current_toolchain = Toolchain::from_settings()?;
+    let maybe_current_toolchain = Toolchain::from_settings();
 
     let toolchain = match toolchain {
         Some(toolchain) => toolchain,
         None => {
             let mut result = String::new();
             if let Some(to) = ToolchainOverride::from_project_root() {
-                let name = match DistToolchainDescription::from_str(&to.cfg.toolchain.channel) {
+                let name = match DistToolchainDescription::from_str(&to.cfg.toolchain.channel.name)
+                {
                     Ok(desc) => desc.to_string(),
-                    Err(_) => to.cfg.toolchain.channel,
+                    Err(_) => to.cfg.toolchain.channel.to_string(),
                 };
                 result.push_str(&format!("{} (override), ", name))
             }
-            result.push_str(&format!("{} (default)", current_toolchain.name));
+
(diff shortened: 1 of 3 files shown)


kayagokalp (approved):
utACK

PR #344, opened 2023-01-16
Fix crash when `.fuelup` doesn't exist

IGI-111:
`write_settings` assumes its parent directory exists when writing
`settings.toml` which is not guaranteed.
This results in a crash when invoking `fuelup toolchain install latest`
if there is no existing `.fuelup`


bingcicle (approved):
Thanks! @IGI-111

PR #349, opened 2023-01-18
refactor: break up command.rs into files per subcommand

### LINKED ISSUE #342: tests: re-org by command type
So far all our 'integration' tests are all in 1 file: `tests/commands.rs`. This has worked well so far but its starting to get a bit messy especially when adding or removing tests.

Now that there are more features and the test suite is more mature, we might want to split `tests/commands.rs` into separate files, eg. we can move all the toolchain related tests to a file named `tests/toolchain.rs`. 

Reference: https://matklad.github.io/2021/05/31/how-to-test.html

deekerno:
Closes #342.

## Changelog
- Break up `command.rs` testing file into separate files organized by subcommand


deekerno:
@bingcicle How do you want me to resolve the lints? I could put some `#[allow(...)]` lints in places, but I want to make sure that I do what's right for the project.

bingcicle:
@deekerno First of all, thanks for the PR! :smile: 

It seems like declaring testcfg as public fixes the issue: https://github.com/rust-lang/rust/issues/46379#issuecomment-548787629. Not sure if there's a better way to go about this, but this looks like it will do for now!

tests/utils/mod.rs
```diff
@@ -0,0 +1,13 @@
+use std::path::Path;
```
bingcicle, line 1: Hmm, I'm personally not really a fan of naming things `utils`, without enough care over time it feels like pieces of code that get hard to categorize turn into _utils_ and it just becomes a dumping ground of unrelated functions, perhaps we can name this something else? eg. `checks`, or `expects`

bingcicle (changes requested):

bingcicle (approved):
LGTM :tada: thanks for the PR!

deekerno:
@bingcicle Looks like it's failing on the same test that master is currently failing on. 🤔

PR #354, opened 2023-01-24
fix: install overridden components to store instead of to the toolchain dir

### LINKED ISSUE #353: bug: installing from override does not work
Reported by @Braqzen:

With `fuelup v0.16.0`:

1. `forc new stst`
2. `cd stst`
3. use the below toml:

```toml
[toolchain]
channel = "latest-2023-01-17"

[components]
forc = "0.33.1"
fuel-core = "0.15.3"
```

4. `forc`

`forc` was not downloaded to the store.

The correct behaviour should mean that `forc` is downloaded to the store and the downloaded components linked to the toolchain directory.

bingcicle:
closes #353 

Logic in `proxy_cli.rs` was broken - previously, installation of a toolchain worked as it did in the past, but in the new store paradigm, we have to install components to the store and then put symlinks in the toolchain directory instead. This PR implements this change.

### DIFF
--- a/src/download.rs
+++ b/src/download.rs
@@ -282,6 +282,15 @@ pub fn download_file_and_unpack(download_cfg: &DownloadCfg, dst_dir_path: &Path)
     Ok(())
 }
 
+pub fn link_to_toolchain(toolchain_dir: PathBuf, bins: Vec<PathBuf>) -> Result<()> {
+    for path in bins {
+        if let Some(file_name) = path.file_name() {
+            hard_or_symlink_file(&path, &toolchain_dir.join(file_name))?;
+        }
+    }
+    Ok(())
+}
+
 pub fn link_to_fuelup(bins: Vec<PathBuf>) -> Result<()> {
     let fuelup_bin_path = fuelup_bin();
     for path in bins {
--- a/src/download.rs
+++ b/src/download.rs
@@ -282,11 +282,9 @@ pub fn download_file_and_unpack(download_cfg: &DownloadCfg, dst_dir_path: &Path)
     Ok(())
 }
 
-pub fn link_to_toolchain(toolchain_dir: PathBuf, bins: Vec<PathBuf>) -> Result<()> {
-    for path in bins {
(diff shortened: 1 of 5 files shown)


src/download.rs
```diff
@@ -282,6 +282,15 @@ pub fn download_file_and_unpack(download_cfg: &DownloadCfg, dst_dir_path: &Path)
     Ok(())
 }
 
+pub fn link_to_toolchain(toolchain_dir: PathBuf, bins: Vec<PathBuf>) -> Result<()> {
```
mitchmindtree, line 285:
```suggestion
pub fn link_to_toolchain(toolchain_dir: &Path, bins: Vec<PathBuf>) -> Result<()> {
```

`toolchain_dir` is never used as a `PathBuf` in this function

src/store.rs
```diff
@@ -38,28 +36,32 @@ impl Store {
...
+        // We ensure that component_dir exists above, so its parent must exist here.
+        unpack_bins(&component_dir, &component_dir)?;
+
+        Ok(())
+    }
+
     // This function installs a component into a directory within '/.fuelup/store'.
     // The directory is named '<component_name>-<version>', eg. 'fuel-core-0.15.1'.
-    pub(crate) fn install_component(
+    pub(crate) fn install_toolchain_component(
```
mitchmindtree, line 52: Maybe it's better that the `Store` doesn't know anything about `toolchain`s, and just handles `components`? We could leave it up to the `toolchain` to create its own links. This might require the `install_component` method returning some info about what was installed (e.g. the resulting path). Will leave it up to you if you want to make this change though!

mitchmindtree (dismissed):
Nice! Left a couple suggestions, though no blockers.

mitchmindtree (dismissed):

bingcicle:
@mitchmindtree sorry, this should be good to go now

bingcicle:
@mitchmindtree hang on, need to handle `forc`'s case. Converting to draft first

mitchmindtree (approved):

PR #357, opened 2023-01-24
relink fuelup to components to push updates

### LINKED ISSUE #356: bug: fuelup's updates are not pushed upon `self update`
`fuelup self update` updates the `fuelup` binary itself but not the existing `fuelup` clones in the form of `forc`, `fuel-core`, and other related binaries in `~/.fuelup/bin`.

An example to understand this: `fuelup` handles overrides differently on execution in `v0.16.0`. This requires the fake `forc` within `~/.fuelup/bin` to be able to handle the `v0.16.0` version of `fuelup` code, to be able to detect that its version is not the one required from the override and handle code in the new store paradigm.

bingcicle:
closes #356 

We try to re-link `fuelup` to whatever components exist in `~/.fuelup/bin` at the time of `fuelup self update`.

To test:

1. checkout an older version of fuelup (v0.15 works).
2. build it and move it to replace the version in ~/.fuelup/bin
3. fuelup should not work with `fuel-toolchain.toml` now, you can confirm this by running `fuelup show` with an override in the CWD, and run `forc` (which should run your default `forc`)
4. now, checkout this branch, run `cargo run self update` which should re-link the components.
5. executing `forc` should now run what is in the override, and `fuelup show` shows your override.

### DIFF
--- a/src/ops/fuelup_self.rs
+++ b/src/ops/fuelup_self.rs
@@ -1,12 +1,13 @@
 use anyhow::{bail, Result};
-use component;
+use component::{self, Components};
 use std::{fs, path::Path};
 use tempfile;
 use tracing::{error, info};
 
 use crate::{
     download::{download_file_and_unpack, unpack_bins, DownloadCfg},
-    path::fuelup_bin,
+    file::hard_or_symlink_file,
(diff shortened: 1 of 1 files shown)


mitchmindtree (approved):
utACK

PR #372, opened 2023-01-30
feat: `add_component` adds to store

### LINKED ISSUE #329: feat: adding components should install to the store
Part of the introduction of the store also means we can download all binaries to ~/.fuelup/store.

During `fuelup component add`:

1. check first if the specified component (version included) exists in the store.
2. if exists, simply link the component to the toolchain, i.e. equivalent of doing `ln ~/.fuelup/store/forc-0.31.0/forc ~/.fuelup/toolchains/my-toolchain/bin/forc`, if our custom toolchain is `my-toolchain` for example.
3. if not, download the component to the store, and link from there.

### LINKED ISSUE #352: feat: custom toolchain components should be symlinks to components in the store
Prior to fuelup v0.16.0, we did not have the store where we downloaded all binaries to.

Now we can download all binaries to the store, and a custom toolchain directory can simply be a set of symlinks to the components within the store.

Some advantages:

1) No need to re-download any versions of binaries if you already have them locally
2) All binaries within a single `store` instead of being distributed across different folders

bingcicle:
closes #329, closes #352

incidentally because `add_component()` is re-used in many commands eg. `toolchain install`, `update`, we get the nice effect of this change everywhere!

eg. `fuelup toolchain install latest` will now symlink components from store if they're already there. No need to wait to re-download components again.

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -11,7 +11,7 @@ use tracing::{error, info};
 use crate::channel::{self, is_beta_toolchain, Channel};
 use crate::config::Config;
 use crate::constants::DATE_FORMAT;
-use crate::download::{download_file_and_unpack, link_to_fuelup, unpack_bins, DownloadCfg};
+use crate::download::DownloadCfg;
 use crate::file::hard_or_symlink_file;
 use crate::ops::fuelup_self::self_update;
 use crate::path::{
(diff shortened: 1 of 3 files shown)


src/toolchain.rs
```diff
@@ -228,46 +228,75 @@ impl Toolchain {
...
+                            info!("Fetching core forc dependencies");
+                            if Command::new(forc_bin_path)
+                                .args(["check", "--path", temp_project_path])
+                                .stdout(std::process::Stdio::null())
+                                .status()
+                                .is_err()
+                            {
+                                error!("Failed to fetch core forc dependencies");
+                            };
+                        };
```
mitchmindtree, line 276: Unrelated to this PR, but could be nice to abstract this into a function, i.e. `fn cache_sway_std_libs`

mitchmindtree (approved):
:ok_hand: utACK :ship:

src/toolchain.rs
bingcicle: Yup, I was going to put that here but it made the diff ugly so I pulled it out in https://github.com/FuelLabs/fuelup/pull/372/commits/c893d718a51e70738c8dc8f3a8f9ea7df4c1c6ee :laughing:

PR #388, opened 2023-02-09
ci: build sway-libs

bingcicle:
This is probably why the compatibility check might be failing. The `test.toml` has a dependency, hence the "failed to source dependency" error message here: https://github.com/FuelLabs/fuelup/actions/runs/4124188987/jobs/7126597387

### DIFF
--- a/.github/workflows/test-toolchain-compatibility.yml
+++ b/.github/workflows/test-toolchain-compatibility.yml
@@ -50,6 +50,12 @@ jobs:
       
       - uses: Swatinem/rust-cache@v2
 
+      - name: Install Forc
+        run: cargo install --locked --debug --path ./forc
+
+      - name: Build sway-lib-core and sway-lib-std
+        run: forc build --path sway-lib-core && forc build --path sway-lib-std
+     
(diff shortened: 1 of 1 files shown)


Voxelot (approved):

Voxelot:
Maybe we should setup a manual trigger for this workflow to allow for testing outside of the normal flow on master?

sdankel (approved):

bingcicle:
> Maybe we should setup a manual trigger for this workflow to allow for testing outside of the normal flow on master?

If by testing you mean triggering the tests manually, the workflow is declared for workflow dispatch: https://github.com/FuelLabs/fuelup/actions/workflows/test-toolchain-compatibility.yml

You have to provide the input in a weird json format though, to tell the CI the combinations that the workflow has to test, for eg. the input for that failed run is: `{"job":[{"forc-version":"0.35.0","fuel-core-version":"0.17.0"},{"forc-version":"0.35.0","fuel-core-version":"0.16.1"},{"forc-version":"0.33.1","fuel-core-version":"0.16.1"}]}`

PR #389, opened 2023-02-10
ci: cd into s way when testing

bingcicle:
related: https://github.com/FuelLabs/fuelup/pull/388 

The PR above fixes not being able to source dependency, but the [Forc.toml for the storage purity test](https://github.com/FuelLabs/sway/blob/9a00b5ab67ce67c5f525bdc332f15d6b6e06eb38/test/src/e2e_vm_tests/test_programs/should_fail/storage_clear_incorrect_purity_annotation/Forc.toml) cds an additional level up to the `sway` repo. Fixing this here instead by creating a dedicated sway repo when running e2e tests

edit: maybe this is the cause actually, instead of the build step above.

### DIFF
--- a/.github/workflows/test-toolchain-compatibility.yml
+++ b/.github/workflows/test-toolchain-compatibility.yml
@@ -40,7 +40,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: fuellabs/sway
-          path: . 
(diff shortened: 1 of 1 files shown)


mitchmindtree (approved):
Lgtm

PR #410, opened 2023-03-03
feat: support fetching and showing fuels_version in fuelup show

### LINKED ISSUE #289: Provide a way to view SDK version compatibility for official toolchains
While `fuelup` cannot *ship* the SDK per se, it would be nice if we could at least provide a way of determining a version that is definitely compatible with a toolchain's version of `forc` and `fuel-core`.

We might do this by adding a special step that inspects the version of `fuels` used within the sway repo E2E tests when checking for a new `forc` version?

We can likely only achieve this nicely for the official toolchains, as there is nothing stopping a user from constructing a custom toolchain using versions of `forc` and `forc-client` with differing SDK versions for example.

bingcicle:
closes #289

This doesn't really address the issue directly because it shows the compatible `fuels` version on a component level rather than the toolchain level, which I think is probably better.

I was contemplating between dealing with this on the toolchain-level or on the component-level, and it came down to where we wanted the logic for getting the fuels version to live.

### The problem

Ultimately, trying to get the fuels version means we have to unavoidably fetch it from the `Cargo.toml` directly, whether we solve it on the toolchain-level or component-level.

On the toolchain-level, we could publish the info in our channels, but that means that on a component-level we do not know what `fuels` version is compatible.

On the component-level, we could fetch the info directly on the first installation, but that means in some areas of fuelup we have to hardcode the logic to fetch the info directly, which may be ugly (see). However this method is a lot more flexible and we can leverage the store by storing a `fuels_version` textfile that is cached in the component directory, rather than store it in the toolchain directory.

### Solution 
Ultimately I decided to go with shipping this feature using the latter method.

My justification:

- In both circumstances we __have__ to maintain some ugly logic that hard codes the URI of the `Cargo.toml` file to fetch and read the `fuels` version, so either way we would have to maintain such code. If we had chosen to ship this feature on the toolchain-level, we would have this logic in the `build-channel` script anyway. I'd say the upside of shipping this on the backend side means that we don't need new `fuelup` releases in the case that the URI changes, and that could be a strong case for NOT taking this approach.

- The latter method also allows us to show the tested `fuels` version per component instead of per toolchain, which I think is better DevEx for users.

At this point of my thought process I'm not sure if there's any cataclysmic downside to doing it as such instead of the original proposed method. I'm still open minded as to the best approach, so let me know if there's some flaw in my thinking here!

### DIFF
--- a/src/download.rs
+++ b/src/download.rs
@@ -74,7 +74,7 @@ impl DownloadCfg {
         })
     }
 
-    pub fn from_package(name: &str, package: Package) -> Result<Self> {
+    pub fn from_package(name: &str, package: &Package) -> Result<Self> {
         let target = TargetTriple::from_component(name)?;
         let tarball_name = tarball_name(name, &package.version, &target);
         let tarball_url = package.target[&target.to_string()].url.clone();
@@ -82,7 +82,7 @@ impl DownloadCfg {
         Ok(Self {
             name: name.to_string(),
             target,
-            version: package.version,
+            version: package.version.clone(),
             tarball_name,
             tarball_url,
             hash,
@@ -307,6 +307,61 @@ pub fn unpack_bins(dir: &Path, dst_dir: &Path) -> Result<Vec<PathBuf>> {
     Ok(downloaded)
 }
 
+fn fuels_version_from_toml(toml: toml_edit::Document) -> Result<String> {
+    if let Some(deps) = toml.get("dependencies") {
+        if let Some(fuels) = deps.get("fuels") {
+            let version_str = match fuels.as_value() {
+                Some(toml_edit::Value::String(s)) => s.value().to_string(),
+                Some(toml_edit::Value::InlineTable(t)) => t.get("version").map_or_else(
+                    || "".to_string(),
+                    |v| v.as_str().unwrap_or_default().to_string(),
+                ),
+                _ => String::default(),
+            };
+            println!("vs: {}", version_str);
+
+            return Ok(version_str);
+        } else {
+            bail!("'fuels' dependency does not exist");
+        };
+    };
+
+    bail!("the table 'dependencies' does not exist");
+}
+
+pub fn fetch_fuels_version(cfg: &DownloadCfg) -> Result<String> {
+    let url = match cfg.name.as_str() {
+        "forc" => format!(
+            "https://raw.githubusercontent.com/FuelLabs/sway/v{}/test/src/sdk-harness/Cargo.toml",
+            cfg.version
+        ),
+        "forc-wallet" => {
+            format!(
+                "https://raw.githubusercontent.com/FuelLabs/forc-wallet/v{}/Cargo.toml",
+                cfg.version
+            )
+        }
+        _ => bail!("invalid component to fetch fuels version for"),
+    };
+
+    // auto detect http proxy setting.
+    let handle = if let Ok(proxy) = env::var("http_proxy") {
+        ureq::builder()
+            .user_agent("fuelup")
+            .proxy(ureq::Proxy::new(proxy)?)
+            .build()
+    } else {
+        ureq::builder().user_agent("fuelup").build()
+    };
+
+    if let Ok(resp) = handle.get(&url).call() {
+        let cargo_toml = toml_edit::Document::from_str(&resp.into_string()?)?;
+        return Ok(fuels_version_from_toml(cargo_toml)?);
+    }
+
+    bail!("Failed to get fuels version");
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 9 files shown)


bingcicle:
This should be tested, but currently the tests are not aware of the store yet - created #411 to track this and prioritizing that, since that will require changes to the test env setup that doesn't concern this PR as a whole.

eureka-cpu:
Seems reasonable

src/download.rs
```diff
@@ -321,6 +378,34 @@ mod tests {
...
+            "0.1",
+            fuels_version_from_toml(toml_edit::Document::from_str(toml).unwrap()).unwrap()
+        );
+
+        let toml = r#"
+[package]        
+name = "forc"
+
+[dependencies]
+fuels = { version = "0.1", features = ["some-feature"] }
```
sdankel, line 400: 👍

src/store.rs
```diff
@@ -41,9 +45,47 @@ impl Store {
...
         download_file_and_unpack(cfg, &component_dir)?;
         // We ensure that component_dir exists above, so its parent must exist here.
         unpack_bins(&component_dir, &component_dir)
     }
+
+    pub(crate) fn cache_fuels_version(&self, cfg: &DownloadCfg) -> Result<()> {
+        let dirname = component_dirname(&cfg.name, &cfg.version);
+
+        if let Ok(fuels_version) = fetch_fuels_version(cfg) {
+            info!("caching fuels version");
```
sdankel, line 71:
```suggestion
            info!("Caching fuels version at {:?}", self.path().join(dirname).join("fuels_version"));
```

src/store.rs
```diff
@@ -41,9 +45,47 @@ impl Store {
...
+            ensure_dir_exists(&self.path().join(&dirname))?;
+            let fuels_version_path = self.path().join(dirname).join("fuels_version");
```
sdankel, line 73: It'd be good to store "fuels_version" as a constant since it's used in multiple places

sdankel (dismissed):
Couple nits, otherwise LGTM!

tests/show.rs
```diff
@@ -89,6 +90,7 @@ latest-{target} (default)
 "#
         );
         assert!(stdout.contains(expected_stdout));
+        assert!(!stdout.contains("fuels versions"));
```
sdankel, line 93: The heading "fuels versions" is a little confusing to me because I think of `fuels` as the SDK, not forc. Perhaps a better heading would be "Fuel component versions" ?

sdankel: Or even just "Component versions" or "Package versions" since the use is running fuelup, so the versions much be pertaining to fuel.

bingcicle: > The heading "fuels versions" is a little confusing to me because I think of fuels as the SDK

Yeah actually, that is the point! I think there was a past version of `forc` that deviated from `fuels` which resulted in incompatibility between certain SDK methods and forc itself. So the feature introduced in this PR is meant to serve as a sort of 'guide' as to what fuels version the component was tested/built against.

I agree though it might need better naming or more explanation behind what 'fuels versions' mean :thinking:

bingcicle:
I merged the other PR that introduces testing to this PR - apologies but i'm going to have to re-request reviews soon

src/download.rs
```diff
@@ -307,6 +307,63 @@ pub fn unpack_bins(dir: &Path, dst_dir: &Path) -> Result<Vec<PathBuf>> {
...
+        "forc-wallet" => {
+            format!(
+                "https://raw.githubusercontent.com/FuelLabs/forc-wallet/v{}/Cargo.toml",
+                cfg.version
+            )
+        }
+        _ => bail!("invalid component to fetch fuels version for"),
+    };
+
+    let handle = if let Ok(proxy) = env::var("http_proxy") {
```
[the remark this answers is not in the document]
bingcicle: ideally it should be used all the time to give the user some flexibility if end users rely on a proxy to use fuelup. An external contributor actually introduced this in a PR and I think we should probably replicate this whereever necessary in a separate PR

src/download.rs
```diff
@@ -307,6 +307,63 @@ pub fn unpack_bins(dir: &Path, dst_dir: &Path) -> Result<Vec<PathBuf>> {
...
+            bail!("'fuels' dependency does not exist");
+        };
+    };
+
+    bail!("the table 'dependencies' does not exist");
+}
+
+/// Fetches the Cargo.toml of a component in its repository and tries to read the version of
+/// the `fuels` dependency.
+pub fn fetch_fuels_version(cfg: &DownloadCfg) -> Result<String> {
```
mitchmindtree, line 335: It looks like we're fetching fuels versions upon `fuelup show` - is that right?

Rather than fetching the version manually from `fuelup`, I wonder if the `fuels` versions should be determined as an optional field as a part of the channel/component information generated by CI? Perhaps this might lower the number of fetches that needs to happen (and improve response time) when a user runs `fuelup show`?

bingcicle: This is called by `cache_fuels_version()`, which in turn is called in `install_component()` ([link](https://github.com/FuelLabs/fuelup/pull/410/files/f4326257cc185b5ddab7096082b3bbfae2edaa1e#diff-41a991785a2eba02345db5fdcbf15b660f27b2841b8f61a084a7db66e69bf3eeR46-R60)). So `fuels` version is cached only on install.  So actually when `fuelup show` is called we're just reading from the cached version [here](https://github.com/FuelLabs/fuelup/pull/410/files/f4326257cc185b5ddab7096082b3bbfae2edaa1e#diff-adb596276059ced1ef81d909863ca19057da292efe1aa1d90288ad87df021626R143)!

> I wonder if the fuels versions should be determined as an optional field as a part of the channel/component information generated by CI?

I went into this in the PR text actually - my main reasoning for going for something like this instead is so we can get the `fuels` version on a component-level instead of a toolchain-level. The ugly way we have to hardcode the fetching logic is unavoidable and is just a case of if we want to hide it behind building the channels or use it in fuelup itself.

mitchmindtree: Ahh sorry bing, I should have read more closely, thanks for clarifying! Doing so during installation sounds fine :+1:

mitchmindtree (approved):

PR #414, opened 2023-03-23
ci: index other components

### LINKED ISSUE #377: ci: track latest releases for other components
Currently the CI tracks only the sway and fuel-core repos for releases. This should be updated to track the other repos as well. The flow should be something like:

1) Check for **ALL** components for new releases.
2) If `forc` or `fuel-core` have new releases, we run compatibility checks on them, and publish a new `latest` channel **only if necessary**.
3) If other components have new releases, we will **always** publish a new `latest`. (This is what we lack)

bingcicle:
closes #377

Adds an alternative mode to `compare-versions` to check for other components other than `forc` and `fuel-core`. Modes are selected by inputting an arg after running the `compare-versions` Rust program. `compare-versions` is still ran every hour by the fuelup gh CI, and the workflow is edited to run them as appropriate.

You may either input *compatibility* or *rest*:

**compatibility**: 

runs the same check we had as always. This is run prior to the toolchain compatibility tests - no changes here.

**rest**:

runs a check that compares latest actual released versions versus latest indexed versions in our `channel-fuel-latest.toml` TOML file, and executes the build-channel script to build a new channel if there are any new versions detected.

Note that this excludes `forc` and `fuel-core`. If for example, a new `forc-index` version is discovered, the `compare-versions` script will call `build-channel` to build a new latest channel, with forc and fuel-core versions parsed from the current latest indexed versions. **We cannot indiscriminately build the new latest channel with the latest `forc` and `fuel-core` since they may not pass compatibility checks.**

I haven't tested this scenario but there may be a race condition where a new version of `forc-index` is published and overwritten by an older version if the compatibility tests are running at the same time and finishes after the publishing is done from this seperate check, but IMO this is inconsequential since a new check will commence an hour later anyway.

My only concern here is potentially being rate-limited by GH CI since there's quite a number of components here, but let's see. The CI fails now and then but so far I've ignored it, because it isn't really critical that we missed 1 check every 24 hours when the check runs every hour anyway.

### DIFF
--- a/ci/compare-versions/src/main.rs
+++ b/ci/compare-versions/src/main.rs
@@ -15,9 +15,10 @@
 //! we know are compatible, so we can update the channel if necessary.
 
 use anyhow::{bail, Result};
+use component::Components;
 use semver::Version;
 use serde::{Deserialize, Serialize};
-use std::{io::Read, str::FromStr};
+use std::{io::Read, os::unix::process::CommandExt, process::Command, str::FromStr};
 use toml_edit::Document;
 
 const GITHUB_API_REPOS_BASE_URL: &str = "https://api.github.com/repos/FuelLabs/";
@@ -30,6 +31,13 @@ const CHANNEL_FUEL_LATEST_TOML_URL: &str =
 #[derive(Debug, Serialize, Deserialize)]
 struct WorkflowRunApiResponse {
     workflow_runs: Vec<WorkflowRun>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct LatestReleaseApiResponse {
+    url: String,
+    tag_name: String,
+    name: String,
 }
 
 #[derive(Debug, Serialize, Deserialize)]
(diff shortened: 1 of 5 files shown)


ci/compare-versions/src/main.rs
```diff
@@ -129,7 +160,74 @@ fn print_selected_versions<'a>(
...
+
+    let channel_doc = toml_resp
+        .parse::<Document>()
+        .expect("invalid channel.toml parsed");
+
+    let components = Components::collect_publishables()?;
+
+    for component in components
+        .iter()
+        .filter(|c| !["forc", "fuel-core"].contains(&c.name.as_str()))
```
deekerno, line 186: `"fuel-indexer"` should be added here as we maintain both `fuel-indexer` and  `forc-index`.

deekerno (changes requested):

ci/compare-versions/src/main.rs
bingcicle: Should `fuel-indexer` not be a downloadable component?

deekerno: Oh geez, I read it wrong. So sorry about that. 😓

deekerno (dismissed):

rvmelkonian (changes requested):
This PR is definitely needed, thanks for putting it up. I think it's likely we will get rate limited, so we just have to make sure any failure cases are being handled gracefully. I've added just one suggestion for error handling.

ci/compare-versions/src/main.rs
```diff
@@ -73,6 +81,29 @@ fn get_workflow_runs(repo: &str) -> Result<WorkflowRunApiResponse> {
...
+        "https://api.github.com/repos/FuelLabs/{}/releases/latest",
+        repo
+    );
+
+    let handle = ureq::builder().user_agent("fuelup").build();
+    let resp = handle
+        .get(&url)
+        .call()
+        .expect(&format!("Could not get latest release for {}", repo));
+
```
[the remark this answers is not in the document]
bingcicle: I opted to go for a match call instead here

deekerno (approved):

rvmelkonian (approved):

PR #419, opened 2023-04-03
docs: explain how fuelup works, add store

### LINKED ISSUE #415: docs: add an architecture page in the book
This should explain:

- proxying between 'fake' binaries in `~/.fuelup/bin` to the real binaries in the toolchain
- the store

bingcicle:
closes #415

Add a `how fuelup works` page (instead of architecture to be more specific) that explains on a high-level and summarizes what the 'binaries' within `~/.fuelup/bin` are, and briefly talks about how it proxies the real binaries.

Also added a page briefly explaining the store.

Would be helpful to review based on whether the explanation helps from an external dev looking for info on how fuelup works.


kayagokalp (approved):
lgtm

PR #423, opened 2023-04-05
fix: proxy_cli erroneously executes `forc` when forc plugins are called

### LINKED ISSUE #422: `forc deploy` binary not found when fuel-toolchain.toml is present
### Problem

When inside a directory with a fuel-toolchain.toml ([example](https://github.com/sarahschwartz/sway-farm/tree/main/contract)), `fuel deploy --help` executes the `forc` executable rather than `forc-deploy`

If I use the full path to the executable, i.e. `$HOME/.fuelup/toolchains/beta-3-aarch64-apple-darwin/bin/forc-deploy --help`, it executes `forc-deploy` correctly. However, if I use `$HOME/.fuelup/bin/forc-deploy --help`, it does not work.

### Steps

If I use the full path to the executable, i.e. `$HOME/.fuelup/toolchains/beta-3-aarch64-apple-darwin/bin/forc-deploy --help`, it executes `forc-deploy` correctly. However, if I use `$HOME/.fuelup/bin/forc-deploy --help`, it does not work.

### Possible Solution(s)

_No response_

### Notes

_No response_

### Fuelup version

```console
fuelup 0.18.1
```

### Installed components

```console
beta-3-aarch64-apple-darwin
  forc - Up to date : 0.35.5
  fuel-core - Up to date : 0.17.4
  fuel-indexer - Up to date : 0.6.0
fuelup - Up to date : 0.18.1
```

bingcicle:
closes #422 

The entire bug was a result of a one-line mistake:

https://github.com/FuelLabs/fuelup/blob/784edcddd8dc28b7499b4410d5c3fc8802ebe281/src/proxy_cli.rs#L74

If you look at the source code you can see that `component_name` is used to derive the path of a binary if a version was specified in an override. In the case of forc, we require this `component_name` instead of using `proc_name` as seen in the function because forc plugins are packaged together with forc - which means if we want to find  the directory that contains `forc-deploy v0.35.5`, this would be found in `forc-0.35.5`, which means we would have to use `component_name` (forc) instead of `proc_name` (forc-deploy, for example)

The bug was caused by using the `component_name` to invoke the binary instead of using the `proc_name` in the case that an entire toolchain was overridden. In that scenario, since that specific version is already linked to the toolchain directory, we do not have to derive the path, we just have to call what's inside the toolchain bin directory, i.e. `~/.fuelup/toolchains/beta-3.../bin/forc-deploy`.

I also added a simple test in this PR to emulate the scenario with a custom injected override file so we can hopefully prevent any regressions on this behavior.

### DIFF
--- a/tests/override.rs
+++ b/tests/override.rs
@@ -9,12 +9,12 @@ use fuelup::{
 use testcfg::FuelupState;
 
 #[test]
-fn check_correct_forc_deploy_called() -> Result<()> {
+fn check_correct_forc_plugin_called() -> Result<()> {
     testcfg::setup(FuelupState::AllInstalled, &|cfg| {
-        let mut stdout = cfg.forc(&["deploy", "--version"]).stdout;
-        assert_eq!(stdout, "forc-deploy 0.1.0\n");
-        stdout = cfg.exec("forc-deploy", &["--version"]).stdout;
-        assert_eq!(stdout, "forc-deploy 0.1.0\n");
+        let mut stdout = cfg.forc(&["wallet", "--version"]).stdout;
+        assert_eq!(stdout, "forc-wallet 0.1.0\n");
+        stdout = cfg.exec("forc-wallet", &["--version"]).stdout;
+        assert_eq!(stdout, "forc-wallet 0.1.0\n");
 
         let toolchain_override = ToolchainOverride {
             cfg: OverrideCfg::new(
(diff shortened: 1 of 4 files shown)


tests/override.rs
```diff
@@ -0,0 +1,42 @@
...
+
+pub mod testcfg;
+use fuelup::{
+    constants::FUEL_TOOLCHAIN_TOML_FILE,
+    toolchain_override::{self, OverrideCfg, ToolchainCfg, ToolchainOverride},
+};
+use testcfg::FuelupState;
+
+#[test]
+fn check_correct_forc_plugin_called() -> Result<()> {
```
sdankel, line 12: 🏆

PR #431, opened 2023-04-10
chore: deprecate usage of hashes to check download duplication

### LINKED ISSUE #418: deprecate checking hashes to skip inefficient redownloads
Previously, we used hashes as a way to detect if a toolchain has been downloaded to tell the user that a toolchain has already been downloaded:

https://github.com/FuelLabs/fuelup/blob/3fa4f390ab916d7d1ba7b8cadc8a79920bffb150/src/toolchain.rs#L328-L337

With the move towards the store-based setup and per-component downloads, the components are 'cached' within the store. Trying to download a toolchain now checks the store for an existing component before attempting to fetch from the channel, which makes the download process a lot quicker if you had already downloaded a component before. Therefore it seems to me there isn't a reason to check hashes anymore.

bingcicle:
Closes #418 

Prior to the 'store' model that we now use, each binary was not cached in some local location. We used basic sha256 hashes stored within .fuelup/hashes directory as a way to tell fuelup that a toolchain already exists. This is important to ensure that users don't waste time re-downloading binaries that they have already downloaded once before.

With the store model, we no longer need this since it's much faster now to re-install toolchains with duplicate binaries since they are all cached in the store. We can now get rid of this hacky way of speeding up the download process, since it's built-in to the design.

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -11,7 +11,6 @@ use anyhow::{bail, Result};
 use component::Components;
 use semver::Version;
 use serde::{Deserialize, Serialize};
-use sha2::{Digest, Sha256};
 use std::{collections::BTreeMap, fmt::Debug};
 use time::Date;
 use toml_edit::de;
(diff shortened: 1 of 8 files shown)


bingcicle:
Note that this actually removes the summary message that says that the toolchain is unchanged since fuelup checks for the hashes for this functionality - I will open a followup PR to address this.

JoshuaBatty (approved):

PR #443, opened 2023-05-23
fix: use forc to build sdk-harness tests in CI

kayagokalp:
Sway repo uses forc workspaces to build sdk harness tests but fuelup CI was still trying to use outdated `buil.sh` which does not exist anymore. So the integration tests were failing, this should fix it but I am not sure how to test it.


mitchmindtree (approved):
Good catch @kayagokalp, lgtm!

PR #449, opened 2023-06-20
docs: fix broken links

sarahschwartz:


kayagokalp (approved):
utAck 💯  We should probably add broken link checking step to CI for here as well. Also would linking to latest versions of the books (instead of master version) make sense?  Just renamed the PR name to match conventional commits as we try to use that in tooling repos

PR #461, opened 2023-07-20
feat: Adds `nix` subcommand that uses the `fuel.nix` flake to manage binaries

### LINKED ISSUE #457: Use fuel.nix internally and deprecate current work.
We have decided to focus our efforts for managing the fuel toolchain around [fuel.nix](https://github.com/FuelLabs/fuel.nix/)

As such, we would like to deprecate the existing internals of `fuelup` and instead wrap the current CLI to call into `fuel.nix` directly. Although users are welcome to use fuel.nix directly, there are a large percentage of users who will still appreciate the fuelup CLI interface for it's simplicity and don't care how the toolchain is managed behind the hood.

eureka-cpu:
Adds a `fuelup nix` subcommand that makes calls into the `fuel.nix` flake. This does not break the current `fuelup`, but will ease users into trying out the nix approach to managing toolchains and components.

Closes #457 

# Process for achieving completion of #457

- [x] In `fuelup-init.sh` use `check_cmd` for `nix`, if it is installed for the user there is no need to run the installation script.
   - [x] ~find a way to amend their `conf.nix` file to include the fuel binary cache from cachix and the necessary unstable features (also check if those features are already enabled)~ out of the scope of this PR, handled by way of ending message
- [x] If `nix` isn't present, run the full `fuel.nix` install script.
- [ ] Create a `fuelup` profile to separate fuel binaries from existing users main profile (this can be a follow up issue)
- [x] Map `fuel.nix` flake commands to the fuelup CLI commands already available.
   - [x] `fuelup nix install`: install toolchain and components
      - [x] auto prioritize the newly installed package
      - [x] print child stderr
      - [ ] add verbosity flag
      - [ ] remove old toolchain or component before installing to avoid duplicate binaries with different priorities (reliant on updates to `fuelup nix list`)
   - [x] `fuelup nix list`: list installed toolchains and components (nix default)
      - [ ] make `fuelup nix list` more robust by organizing the display of toolchains and components in a more similar way to the current fuelup 
   - [x] `fuelup nix remove`: remove toolchains and components
   - [x] `fuelup nix upgrade`: upgrade toolchains and components
   - enable nix shell #463 
- [x] Run tests to ensure all features work as intended.
   - [x] test locally
   - [x] unit tests where applicable
   - [x] CI tests where applicable


JoshuaBatty:
Nice work so far @eureka-cpu . 

Can we copy over the appropriate parts from the fuel.nix CI so we can test that all the features?

eureka-cpu:
> Nice work so far @eureka-cpu .
> 
> Can we copy over the appropriate parts from the fuel.nix CI so we can test that all the features?

@JoshuaBatty I've got the fuelup nix commands working in CI but there seems to be an issue with fuelup check on master #465

eureka-cpu:
@JoshuaBatty 

Here's the 3 ways I've tested nix installations with the flake:
- Install nix normally then add necessary extra features
- Install nix from the flake link
- Use a nixOS system and add the necessary extra features

All three methods take a long time to build, just like here in CI

JoshuaBatty:
I'm trying to test this out now. I pulled down the PR and installed, made sure there were no other versions of fuelup on my system. 

I then ran:

```
fuelup % fuelup nix install latest
downloading and installing fuel latest toolchain, this may take a while...
```

Now i'm just left hanging in a terminal with the only UX message stating that this will take a while. It's pretty hard to understand what is happening behind the scenes. 

Is there a verbose flag I can add to see what is happening behind the scenes?

JoshuaBatty:
I'm not sure if this is within the scope of this PR but I would expect a sort of TUI to display once this command has executed showing the following status:

- what is being downloaded, 
- how far along the installation process are we
- estimated time remaining.

Not being able to see this information leads to a really bad UX.

eureka-cpu:
> I'm not sure if this is within the scope of this PR but I would expect a sort of TUI to display once this command has executed showing the following status:
> 
> * what is being downloaded
> * how far along the installation process are we
> * estimated time remaining.
> 
> Not being able to see this information leads to a really bad UX.

Unfortunately, unless there's a better crate or we want to do it a less rusty way, the `std::process::Command` `output` is the only way to get `std_out` from a `Command` and it is only available after the command has finished which is why you'll see everything that you'd normally see during the install phase, at the end.

I could write a shell script that will try to run the nix commands instead, similarly to how I am doing the `fuelup nix shell` command, but that would take some time. Let me know if that approach works better. We would definitely have better UX that way, it would just be the standard nix output but it's better than only seeing output after the command has executed which can take a long time (hence the message).

eureka-cpu:
> I'm trying to test this out now. I pulled down the PR and installed, made sure there were no other versions of fuelup on my system.
> 
> I then ran:
> 
> ```
> fuelup % fuelup nix install latest
> downloading and installing fuel latest toolchain, this may take a while...
> ```
> 
> Now i'm just left hanging in a terminal with the only UX message stating that this will take a while. It's pretty hard to understand what is happening behind the scenes.
> 
> Is there a verbose flag I can add to see what is happening behind the scenes?

Unfortunately as stated before, there's not much we can do about that since the command takes a long time to finish and we don't get access to `std_out` until it finishes.

JoshuaBatty:
> Unfortunately, unless there's a better crate or we want to do it a less rusty way, the `std::process::Command` `output` is the only way to get `std_out` from a `Command` and it is only available after the command has finished which is why you'll see everything that you'd normally see during the install phase, at the end.

Instead of calling `output()` you can pipe the child processes output to the user's terminal by doing the following.

```
Command::new(NIX_CMD)
    .stdout(Stdio::inherit())
    .spawn();
```

eureka-cpu:
> > Unfortunately, unless there's a better crate or we want to do it a less rusty way, the `std::process::Command` `output` is the only way to get `std_out` from a `Command` and it is only available after the command has finished which is why you'll see everything that you'd normally see during the install phase, at the end.
> 
> Instead of calling `output()` you can pipe the child processes output to the user's terminal by doing the following.
> 
> ```
> Command::new(NIX_CMD)
>     .stdout(Stdio::inherit())
>     .spawn();
> ```

handy, thanks! I read the docs but didn't see that part

JoshuaBatty:
[This guide](https://docs.cachix.org/faq#why-is-nix-not-picking-up-on-any-of-the-pre-built-artifacts) looks like it might have some handy tips to debug why the binary caches aren't being used.

JoshuaBatty:
Leaving some more UX feedback. 

After installing and running `fuelup nix` i get the following:

```
fuelup % fuelup nix
Use the fuel.nix flake to manage toolchains and components. Refer to the fuel.nix book for more info: https://github.com/FuelLabs/fuel.nix/blob/master/book/src/packages.md#packages

Usage: fuelup nix <COMMAND>

Commands:
  install  Install a distributable toolchain or component
  remove   Uninstall a toolchain or component by providing its index, unlocked attribute path or nix store path
  upgrade  Upgrade installed packages by index or unlocked attribute path with the latest version of the fuel.nix flake. Upgrades all installed packages if no index or path is provided
  list     Lists the installed packages by index, unlocked attribute path, locked attribute path and nix store path, respectively
  help     Print this message or the help of the given subcommand(s)
  ```

Can we remove this part below and just show the usage...

```
Use the fuel.nix flake to manage toolchains and components. Refer to the fuel.nix book for more info: https://github.com/FuelLabs/fuel.nix/blob/master/book/src/packages.md#packages
```

**Comment by JoshuaBatty** [comment]
I'm running `fuelup nix install latest` with the latest changes.

```
fuelup % fuelup nix install latest
downloading and installing fuel latest toolchain, this may take a while...
copying path '/nix/store/jskppgw1xjb1n7bkf6nmmqw6kk2a0c9z-sway-04a5970' from 'https://fuellabs.cachix.org'...
this derivation will be built:
  /nix/store/yds5gwpd5hhglxrzi3mh1pf6m43ipkyb-fuel-latest.drv
these 6 paths will be fetched (86.43 MiB download, 182.03 MiB unpacked):
  /nix/store/r1mwwpd3wjlwgs00h8pzngdwxk9ibl2m-forc-0.44.1
  /nix/store/161rs18wjmpx0vlvgzskc59c4lj8vwv7-forc-client-0.44.1
  /nix/store/wizcks47lf3zzdgfj94h0fs48qkn0bgh-forc-doc-0.44.1
  /nix/store/iagvlp5vlaycw9qgc89v75irr2xr2bb6-forc-fmt-0.44.1
  ```
  
Although I think printing every line to the console is useful, this should be behind a --verbose flag. By default, only one line is used for the installation status and subsequent messages overwrite the single line in the terminal.

JoshuaBatty:
I also got this message printing to the console

```
building '/nix/store/yds5gwpd5hhglxrzi3mh1pf6m43ipkyb-fuel-latest.drv'...
error: An existing package already provides the following file:

         /nix/store/d2f5baagzqim9flchiswcpf7z8061h6p-fuel-latest/bin/forc-doc

       This is the conflicting file from the new package:

         /nix/store/ib3lyc87qikp98gx9djd2j6354vnjr7c-fuel-latest/bin/forc-doc

       To remove the existing package:

         nix profile remove github:fuellabs/fuel.nix

       The new package can also be installed next to the existing one by assigning a different priority.
       To prioritise the new package:

         nix profile install github:fuellabs/fuel.nix --priority 4

       To prioritise the existing package:

         nix profile install github:fuellabs/fuel.nix --priority 6
         ```
         
 Not sure why this message still got through? 
 
 I then ran the command again and this time it just returned to the console with no status message if the installation was successful or had failed..
 
 ```
 fuelup % fuelup nix install latest
downloading and installing fuel latest toolchain, this may take a while...
 fuelup %
```

eureka-cpu:
> Leaving some more UX feedback.
> 
> After installing and running `fuelup nix` i get the following:
> 
> ```
> fuelup % fuelup nix
> Use the fuel.nix flake to manage toolchains and components. Refer to the fuel.nix book for more info: https://github.com/FuelLabs/fuel.nix/blob/master/book/src/packages.md#packages
> 
> Usage: fuelup nix <COMMAND>
> 
> Commands:
>   install  Install a distributable toolchain or component
>   remove   Uninstall a toolchain or component by providing its index, unlocked attribute path or nix store path
>   upgrade  Upgrade installed packages by index or unlocked attribute path with the latest version of the fuel.nix flake. Upgrades all installed packages if no index or path is provided
>   list     Lists the installed packages by index, unlocked attribute path, locked attribute path and nix store path, respectively
>   help     Print this message or the help of the given subcommand(s)
> ```
> 
> Can we remove this part below and just show the usage...
> 
> ```
> Use the fuel.nix flake to manage toolchains and components. Refer to the fuel.nix book for more info: https://github.com/FuelLabs/fuel.nix/blob/master/book/src/packages.md#packages
> ```
> 
> Also, can we keep the existing command descriptions instead of the ones from `fuel.nix`.

They aren't from the flake, I wrote those because they are accurate to the actual logic behind the command. I can remove them, but it will be less descriptive on how to actually use the commands in some cases

eureka-cpu:
> I also got this message printing to the console
> 
> ```
> building '/nix/store/yds5gwpd5hhglxrzi3mh1pf6m43ipkyb-fuel-latest.drv'...
> error: An existing package already provides the following file:
> 
>          /nix/store/d2f5baagzqim9flchiswcpf7z8061h6p-fuel-latest/bin/forc-doc
> 
>        This is the conflicting file from the new package:
> 
>          /nix/store/ib3lyc87qikp98gx9djd2j6354vnjr7c-fuel-latest/bin/forc-doc
> 
>        To remove the existing package:
> 
>          nix profile remove github:fuellabs/fuel.nix
> 
>        The new package can also be installed next to the existing one by assigning a different priority.
>        To prioritise the new package:
> 
>          nix profile install github:fuellabs/fuel.nix --priority 4
> 
>        To prioritise the existing package:
> 
>          nix profile install github:fuellabs/fuel.nix --priority 6
> ```
> 
> Not sure why this message still got through?
> 
> I then ran the command again and this time it just returned to the console with no status message if the installation was successful or had failed..
> 
> ```
> fuelup % fuelup nix install latest
> downloading and installing fuel latest toolchain, this may take a while...
> fuelup %
> ```

The reason the error was still showing up was because I accidentally removed a line that checked for "error:" in one of my commits where I was cleaning up

the reason you don't see it after if you run it again is because it went through the auto prioritize function

src/commands/nix/install.rs
```diff
@@ -0,0 +1,212 @@
...
+        .arg(link.clone())
+        .output()?;
+    if !output.stderr.is_empty() {
+        let stderr_str = String::from_utf8_lossy(&output.stderr);
+        if stderr_str.contains(NIXOS_PRIORITY_MSG) || stderr_str.contains(NIX_PKG_PRIORITY_MSG) {
+            // recursively decriment the package priority until the
+            // newly installed package has the highest priority
+            try_prioritize(pkg_priority, link)?
+        }
+    }
```
sdankel, line 209: Consider using the rust nix crate instead of the nix CLI: https://docs.rs/nix/latest/nix/

src/commands/nix/install.rs
```diff
@@ -0,0 +1,212 @@
...
+            command.name.clone(),
+            &mut priority_err,
+            &mut all_errs,
+            false,
+        );
+
+        (priority_err.concat(), all_errs.concat(), link)
+    } else {
+        bail!(
+            "available distrubuted components:\n  -fuel-core\n  -fuel-core-client\n  -fuel-indexer\n  -forc\n  -forc-client\n  -forc-doc\n  -forc-explore\n  -forc-fmt\n  -forc-index\n  -forc-lsp\n  -forc-tx\n  -forc-wallet\n  -sway-vim\n
```
sdankel, line 65: Can we pull in this list dynamically rather than hardcoding it?

src/commands/nix/link_generator.rs
```diff
@@ -0,0 +1,207 @@
...
+            FuelComponent::SwayVim => "sway-vim",
+        };
+        let tool = match tool {
+            FuelToolchain::Latest => "",
+            FuelToolchain::Nightly => "-nightly",
+            FuelToolchain::Beta1 => "-beta-1",
+            FuelToolchain::Beta2 => "-beta-2",
+            FuelToolchain::Beta3 => "-beta-3",
+            FuelToolchain::Beta4rc => "-beta-4-rc",
+            FuelToolchain::Unknown => bail!("available distributed toolchains:\n  -latest\n  -nightly\n  -beta-1\n  -beta-2\n  -beta-3\n  -beta-4-rc")
```
kayagokalp, line 60: We also have `beta-4-rc.2` as of #467

src/commands/nix/link_generator.rs
```diff
@@ -0,0 +1,207 @@
...
+#[derive(Eq, PartialEq, Debug)]
+pub(crate) enum FuelToolchain {
+    Latest,
+    Nightly,
+    Beta1,
+    Beta2,
+    Beta3,
+    Beta4rc,
+    Unknown,
+}
```
kayagokalp, line 101: ^ I guess we should also add `Beta4Rc2` to this list as well.

kayagokalp (commented):
Just pointed out some stuff that got merged while this PR was cooking so that if we decide to merge this one we do not lose anything

eureka-cpu:
Thanks @kayagokalp I'll update those soon

src/commands/nix/install.rs
eureka-cpu: Something like this?
```rust
const AVAILABLE_DIST_COMPONENTS: &[FuelComponent; 13] = &[
    FuelComponent::SwayVim,
    // ...
];
impl FuelComponent {
    fn as_str(&self) -> &'static str {
        match self {
            FuelComponent::SwayVim => "-sway-vim",
            // ...
        }
    }
}
// ...
let err_str = AVAILABLE_DIST_COMPONENTS.iter().map(|comp| comp.as_str()).collect::<Vec<&str>>().join('\n');
bail!("available distrubuted components: {err_str}\n")
```

eureka-cpu: I've gone ahead and implemented it this way, lmk if you meant some other way and I can change it.

eureka-cpu:
Closing since we are not using this approach as discussed in #457

PR #467, opened 2023-08-15
feat: add `beta-4-rc-2` dist channel support

kayagokalp:
Adds beta-4-rc2 support to fuelup so that people can install it via:

```console
fuelup toolchain install beta-4-rc.2
```

Unfortunately due to how fuelup parses toolchain metadata it is not straightforward to make it possible to use:

```console
fuelup toolchain install beta-4-rc-2
```

We can support it later on and change the name if necessary as this is more towards to our internal testing, I feel like timing is has a little priority rn, but happy to mark this as draft and handle that if it seems like blocker.


JoshuaBatty:
I think the correct way would be `beta-4-rc.2` . This is the convention rust crates use. Would this be hard to implement in this PR?

kayagokalp:
> I think the correct way would be `beta-4-rc.2` . This is the convention rust crates use. Would this be hard to implement in this PR?

Sounds good! pushed efbcee2, now it should be installing beta-4-rc-2 with `fuelup toolchain install beta-4-rc.2`.

eureka-cpu (approved):

PR #480, opened 2023-08-30
feat: add beta-4 support

kayagokalp:
Adds `beta-4` channel support to fuelup.


Braqzen (approved):
LGTM

Only wondering about the various b-4 releases. That may be confusing for users, no?

JoshuaBatty:
We can probably yank the release candidates in the near future so there is only beta-4 available.

JoshuaBatty (approved):

PR #488, opened 2023-09-19
ci: use bigger image to run check-versions ci workflow

kayagokalp:
Our CI runs are failing, possible due to lack of memory/storage on ci runner. 

This fix is also applied to sway repo ci: https://github.com/FuelLabs/sway/commit/e75f14b03636bc96751a6760304a1a6d3eb5937d


kayagokalp:
We only need to have bigger runner,  where we need to get sway repo in the CI which happens during indexing versions.

sdankel (approved):

PR #490, opened 2023-09-19
Use larger image in test-toolchain-compatibility.yml

sdankel:
This is the reusable workflow where the tests are actually run. I think that's why https://github.com/FuelLabs/fuelup/pull/488 didn't work.

Also adds a manual trigger for this job, and changes the cron to run every 6 hours.


kayagokalp (approved):
🤞

PR #506, opened 2023-11-06
feat: make the latest channel point to beta-4

kayagokalp:
Changes latest channel to point beta-4 which is the latest network. We can later on create a new channel with a different name (maybe `latest-local`?) to offer current latest channel. One other alternative is to keep publishing latest as it is but default to beta-4. I can change the PR to that approach as well according to comments.

### DIFF
--- a/docs/src/concepts/channels.md
+++ b/docs/src/concepts/channels.md
@@ -55,7 +55,7 @@ The `latest` channel is pointing to our latest beta network. This toolchain shou
 
 > **Note**
 >
-> The `latest` channel used to point latest compatible versions of `forc` and `fuel-core`, after version v0.20.0 latest is changed to point latest network. This is a breaking change and should be taken into account for existing workflows.
+> The `latest` channel used to point latest compatible versions of `forc` and `fuel-core`, after version v0.20.0 latest is changed to point to the latest network. This is a breaking change and should be taken into account for existing workflows.
 
 ## Developer Guide
 
--- a/docs/src/concepts/channels.md
+++ b/docs/src/concepts/channels.md
@@ -98,4 +98,3 @@ You may also use [nektos/act](https://github.com/nektos/act) to run the workflow
 [beta-4-rc]: #the-beta-4-rc-channel
 [beta-4-rc2]: #the-beta-4-rc2-channel
 [beta-4]: #the-beta-4-channel
-[gh-pages]: https://github.com/FuelLabs/fuelup/tree/gh-pages
--- a/docs/src/concepts/channels.md
+++ b/docs/src/concepts/channels.md
@@ -6,9 +6,9 @@
 
 | Channel       | Source          | Integration Tested   | Update Frequency         | Available |
 | ------------- | --------------- | -------------------- | ------------------------ | --------- |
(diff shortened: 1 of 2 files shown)


src/constants.rs
```diff
@@ -6,7 +6,7 @@ pub const FUELS_VERSION_FILE: &str = "fuels_version";
 
 pub const CHANNEL_LATEST_URL: &str =
     "https://raw.githubusercontent.com/FuelLabs/fuelup/gh-pages/channel-fuel-latest.toml";
-pub const CHANNEL_LATEST_FILE_NAME: &str = "channel-fuel-latest.toml";
+pub const CHANNEL_LATEST_FILE_NAME: &str = "channel-fuel-beta-4.toml";
```
Voxelot, line 9: Will fuelup update in this case just download any updates to beta4 (such as patch releases)?

Voxelot (dismissed):

src/constants.rs
```diff
@@ -6,7 +6,7 @@ pub const FUELS_VERSION_FILE: &str = "fuels_version";
 
 pub const CHANNEL_LATEST_URL: &str =
     "https://raw.githubusercontent.com/FuelLabs/fuelup/gh-pages/channel-fuel-latest.toml";
```
sdankel, line 8: Should we remove this, and the other code for `latest`? It might be confusing to leave it in there.

sdankel (commented):
Do we need to update docs as well?

src/constants.rs
kayagokalp: Yeah update checks the channel file, since channel file name is now beta-4, it should update correctly.

src/constants.rs
kayagokalp: Nice catch!

sdankel (approved):

docs/src/concepts/channels.md
```diff
@@ -50,17 +50,12 @@ Keep in mind that compatibility between `forc` and `fuel-core` is not guaranteed
...
 
-When installing the `latest` channel, fuelup will refer to the `channel-fuel-latest.toml` file published within the fuelup repository in the [gh-pages] branch to determine the set of `forc` and `fuel-core` versions to retrieve. The versions in this file are updated by a scheduled GitHub workflow that runs once every 30 minutes and performs the following steps:
-
-1. Checks for newly published versions of forc and fuel-core.
-2. Tests compatibility of new versions against a set of integration tests.
-3. Selects the latest set of versions that successfully pass the tests.
-4. Publishes the selected versions to the channel-fuel-latest.toml manifest.
+> **Note**
+>
+> The `latest` channel used to point latest compatible versions of `forc` and `fuel-core`, after version v0.20.0 latest is changed to point latest network. This is a breaking change and should be taken into account for existing workflows.
```
JoshuaBatty, line 58:
```suggestion
> The `latest` channel used to point latest compatible versions of `forc` and `fuel-core`, after version v0.20.0 latest is changed to point to the latest network. This is a breaking change and should be taken into account for existing workflows.
```

JoshuaBatty (changes requested):

JoshuaBatty (approved):

PR #521, opened 2023-11-21
fix: don't build proxy agent when env 'http_proxy' is empty

lilasxie:
Running `fuelup component list` with empty `http_proxy` environment value will cause `failed to get latest version` info.

We should check `http_proxy` environment value before building proxy agent.
```
Installed:
  forc 0.46.1 (latest: failed to get latest version)
    - forc-fmtforc-lspforc-docforc-deployforc-run
  forc-explore 0.28.1 (latest: failed to get latest version)
  forc-index 0.23.0 (latest: failed to get latest version)
  forc-wallet 0.3.0 (latest: failed to get latest version)
  fuel-core 0.20.5 (latest: failed to get latest version)
  fuel-core-keygen : failed getting current version
  fuel-indexer 0.23.0 (latest: failed to get latest version)
```


sdankel:
Could you paste the full command you're using to repro this?

src/download.rs
```diff
@@ -92,7 +92,9 @@ pub fn build_agent() -> Result<ureq::Agent> {
     let agent_builder = ureq::builder().user_agent("fuelup");
 
     if let Ok(proxy) = env::var("http_proxy") {
-        return Ok(agent_builder.proxy(ureq::Proxy::new(proxy)?).build());
+        if !proxy.is_empty() {
+            return Ok(agent_builder.proxy(ureq::Proxy::new(proxy)?).build());
+        }
```
sdankel, line 97: I think it would be better to surface the error to the user, in case they have any invalid value for `http_proxy`

```suggestion
        return match ureq::Proxy::new(proxy) {
            Some(proxy) => Ok(agent_builder.proxy(proxy).build()),
            Err(err) => {
                error!("Failed to build proxy with http_proxy={proxy}");
                debug!("{:?}", err);
                Err(err)
            }
        };
```

lilasxie: > I think it would be better to surface the error to the user, in case they have any invalid value for `http_proxy`

Hi @sdankel the suggestion code will be fail to compile, I have fixed that, thanks for your review *_*

sdankel (approved):

PR #523, opened 2023-11-22
doc: add fuelup http proxy setting document

lilasxie:
As I first run `fuelup toolchain install latest`, I failed because of network errors, and didn't find the instruction to set an http proxy to solve this problem. We should add this part IMO 😀

### DIFF
--- a/docs/src/basics.md
+++ b/docs/src/basics.md
@@ -36,9 +36,17 @@ fuelup self update
 
 ## Using Http Proxy
 
-To configure `fuelup` to use your proxy setting you can change `http_proxy` environment value. The value format is in [libcurl format](https://everything.curl.dev/libcurl/proxies#proxy-types) as in `[protocol://]host[:port]`.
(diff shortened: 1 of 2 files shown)


docs/src/basics.md
```diff
@@ -35,6 +35,12 @@ fuelup self update
 ```
 <!-- update_fuelup:example:end -->
 
+## Using Http Proxy
+
+To configure `fuelup` to use your proxy setting you can change `http_proxy` environment value. The value format is in [libcurl format](https://everything.curl.dev/libcurl/proxies#proxy-types) as in `[protocol://]host[:port]`.
+
+***Warning: don't leave `http_proxy` environment with empty string or other invalid format***
```
kayagokalp, line 41: We can add other options here as well since they are added with #592

lilasxie: @kayagokalp Suggestion changes addressed~ Thanks for your review

kayagokalp (approved):
Thanks a lot!

Issue #182, opened 2022-09-09
create helpers to print lines prefixed with whitespaces


bingcicle:
We have places in `fuelup show` and `fuelup check` where we'd like to show the hierarchy of components/plugins by prefixing a certain amount of whitespaces to stagger the output eg. we show forc-fmt as part of forc, for example. Right now we manually type in the spaces, but we can improve by having a fn to input the number of spaces we want instead.

mitchmindtree:
I think the `format!` macro might already allow for this with one of its specifiers, e.g. https://doc.rust-lang.org/std/fmt/index.html#fillalignment.

PR #524, opened 2023-11-22
feat: using rust's built-in alignment to place spaces we want

levisyin:
close [#182](https://github.com/FuelLabs/fuelup/issues/182)

This PR added a function `pad_spaces(num: usize) -> String` to conveniently input the number of spaces we want.

### DIFF
--- a/src/ops/fuelup_check.rs
+++ b/src/ops/fuelup_check.rs
@@ -66,13 +66,14 @@ fn check_plugin(plugin_executable: &Path, plugin: &str, latest_version: &Version
                 Some(v) => {
                     let version = Version::parse(v)?;
                     info!(
-                        "    - {} - {}",
+                        "{:>4}- {} - {}",
+                        "",
                         bold(plugin),
                         format_version_comparison(&version, latest_version)
                     );
                 }
                 None => {
-                    info!("    - {} - Error getting version string", plugin);
+                    info!("{:>4}- {} - Error getting version string", "", plugin);
                 }
             };
         }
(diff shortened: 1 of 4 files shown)


src/ops/fuelup_check.rs
```diff
@@ -60,18 +60,18 @@ fn check_plugin(plugin_executable: &Path, plugin: &str, latest_version: &Version
             match output.split_whitespace().last() {
                 Some(v) => {
                     let version = Version::parse(v)?;
-                    print!("    - ");
+                    print!("{}- ", pad_spaces(4));
```
JoshuaBatty, line 63:
```suggestion
                    println!("{:>4}- ", "");
```

This seems more idiomatic.

JoshuaBatty:
Thanks for the PR @levisyin! This is certainly nicer than what is currently being done. 

Instead of using the `pad_spaces` function, perhaps we should use rust's built-in alignment capabilities as I suggested above. You can see more info about that [here](https://doc.rust-lang.org/std/fmt/index.html#fillalignment). Thanks!

levisyin:
> Thanks for the PR @levisyin! This is certainly nicer than what is currently being done.
> 
> Instead of using the `pad_spaces` function, perhaps we should use rust's built-in alignment capabilities as I suggested above. You can see more info about that [here](https://doc.rust-lang.org/std/fmt/index.html#fillalignment). Thanks!

Hi sir, using rust's built-in alignment won't work well for issue #182 . Think about this:
I wanna print!("    -hello world"), with 4 space left on "-hello world". If using rust's built-in alignment, we should calculate the whole str length to find how many padding we should set. In this case, string length is 16, so we should set println!("{:>16}- ", ""). This is not convenient enough for us to quick set the num space we wanna leave.

levisyin:
Done! See following creenshot:

image

JoshuaBatty (approved):
Nice, thanks for the PR!

Issue #319, opened 2022-12-18
feat: progression indicators for component/toolchain installation


bingcicle:
It can be useful to have a progress indicator during commands that add/install components/toolchains so that we at least know the executable is still doing work, especially when met with low bandwidth/slow internet speeds.

PR #525, opened 2023-11-23
feat: support display download progress bar

Halimao:
close [#319](https://github.com/FuelLabs/fuelup/issues/319)

Add downloading progress bar to commands that add/install components/toolchains, now we could know wether the executable is still doing work when we met with low bandwidth/slow internet speeds.

See following pictures
image

image


sdankel:
Thanks for this! How does it look in the log output (`~/.fuelup/log/`)? It would be nice to be able to log everything the user sees. [This crate](https://crates.io/crates/tracing-indicatif) may be useful.

src/download.rs
```diff
@@ -203,17 +254,60 @@ pub fn download_file(url: &str, path: &PathBuf) -> Result<()> {
...
+                debug!(
+                    "[{}] [{}] {}/{} {}/s ({}) - {}",
+                    FormattedDuration(progress_bar.elapsed()),
+                    "#".repeat(40),
+                    HumanBytes(progress_bar.position()),
+                    HumanBytes(progress_bar.length().unwrap_or(progress_bar.position())),
+                    HumanBytes(progress_bar.per_sec() as u64),
+                    HumanDuration(progress_bar.eta()),
+                    progress_bar.message(),
+                );
```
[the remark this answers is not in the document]
Halimao: @sdankel Done!

src/download.rs
```diff
@@ -291,6 +285,63 @@ pub fn unpack_bins(dir: &Path, dst_dir: &Path) -> Result<Vec<PathBuf>> {
...
+    let total_size = response
+        .header("Content-Length")
+        .and_then(|s| s.parse::<u64>().ok())
+        .unwrap_or(0);
+    let mut downloaded_size = 0;
+    let mut buffer = [0; 8192];
+    let progress_bar = ProgressBar::new(total_size);
+    progress_bar.set_style(
+                ProgressStyle::default_bar()
+                    .template("[{elapsed_precise}] [{bar:40.cyan/blue}] {bytes}/{total_bytes} {bytes_per_sec} ({eta}) - {msg:.cyan}")
```
sdankel, line 302: I don't think we need to show the bytes per sec, and it's nicer to have "Download complete" in green

```suggestion
                    .template("[{elapsed_precise}] [{bar:40.cyan/blue}] {bytes}/{total_bytes} ({eta}) - {msg:.green}")
```

src/download.rs
```diff
@@ -291,6 +285,63 @@ pub fn unpack_bins(dir: &Path, dst_dir: &Path) -> Result<Vec<PathBuf>> {
...
+                HumanBytes(progress_bar.per_sec() as u64),
+                HumanDuration(progress_bar.eta()),
+                progress_bar.message(),
+            );
+            error!("Something went wrong writing data: {}", e)
+        };
+        downloaded_size += bytes_read as u64;
+        progress_bar.set_position(downloaded_size);
+    }
+    progress_bar.finish_with_message("Download complete");
```
sdankel, line 331: I think this shows that the download completed even if there was an error. Could you add a unit test where it fails?

src/download.rs
```diff
@@ -291,6 +285,63 @@ pub fn unpack_bins(dir: &Path, dst_dir: &Path) -> Result<Vec<PathBuf>> {
...
+                "#".repeat(
+                    (progress_bar.position() * 40 / progress_bar.length().unwrap()) as usize
+                ),
+                HumanBytes(progress_bar.position()),
+                HumanBytes(progress_bar.length().unwrap_or(progress_bar.position())),
+                HumanBytes(progress_bar.per_sec() as u64),
+                HumanDuration(progress_bar.eta()),
+                progress_bar.message(),
+            );
+            error!("Something went wrong writing data: {}", e)
```
sdankel, line 326: Could we still log the path that it failed to write to? I think this should still be a `bail!` so that we exit if the download failed.

src/download.rs
```diff
@@ -291,6 +285,63 @@ pub fn unpack_bins(dir: &Path, dst_dir: &Path) -> Result<Vec<PathBuf>> {
...
+                FormattedDuration(progress_bar.elapsed()),
+                "#".repeat(
+                    (progress_bar.position() * 40 / progress_bar.length().unwrap()) as usize
+                ),
+                HumanBytes(progress_bar.position()),
+                HumanBytes(progress_bar.length().unwrap_or(progress_bar.position())),
+                HumanBytes(progress_bar.per_sec() as u64),
+                HumanDuration(progress_bar.eta()),
+                progress_bar.message(),
+            );
```
sdankel, line 325: Could we move this to a function and call it in both places?

```
fn log_progress_bar(progress_bar: &ProgressBar) {
    debug!(
        "[{}] [{}] {}/{} ({}) - {}",
        FormattedDuration(progress_bar.elapsed()),
        "#".repeat(
            (progress_bar.position() * 40 / progress_bar.length().unwrap()) as usize
        ),
        HumanBytes(progress_bar.position()),
        HumanBytes(progress_bar.length().unwrap_or(progress_bar.position())),
        HumanDuration(progress_bar.eta()),
        progress_bar.message(),
    );
}
```

sdankel (commented):
Tested this out and it's looking good! It makes me think there's a lot more we could do to improve the look and feel of fuelup with spinners and multiline progress bars (if we download in parallel). Left a few comments.

src/download.rs
```diff
@@ -11,8 +12,8 @@ use std::str::FromStr;
 use std::time::Duration;
 use std::{fs, thread};
 use tar::Archive;
-use tracing::warn;
-use tracing::{error, info};
+use tracing::{debug, error, info, warn};
+use ureq::Response;
```
sdankel, line 16: Don't think we need this?

src/download.rs
```diff
@@ -224,6 +216,8 @@ pub fn download_file(url: &str, path: &PathBuf) -> Result<()> {
                 let retry = retry.unwrap_or(RETRY_DELAY_SECS);
                 info!("Retrying..");
                 thread::sleep(Duration::from_secs(retry));
+                // clean file content every retry
+                file = OpenOptions::new().write(true).truncate(true).open(path)?;
```
sdankel, line 220: I don't think we need to do this. If we got a 404, nothing was written.

src/download.rs
Halimao: I have tested in local, when download encount error, won't show "download completed"

src/download.rs
Halimao: Done

src/download.rs
```diff
@@ -480,9 +501,26 @@ fuels = { version = "0.1", features = ["some-feature"] }
...
+        let body = "A".repeat(len);
+        let s = format!(
+            "HTTP/1.1 200 OK\r\n\
+                 Content-Length: {}\r\n
+                 \r\n
+                 {}",
+            len, body,
+        );
+        let res = s.parse::<Response>().unwrap();
+        assert!(write_response_with_progress_bar(res, &mut mock_writer, String::new()).is_ok());
```
Braqzen, line 524: I don't believe that we need to assert or `is_ok()` here, no?
If the call fails then the panic will be triggered and the test will fail as expected, or am I misunderstanding?

Halimao: `anyhow::bail!` doestn't triigger panic
image

Halimao: Or we can remove [should_panic], just use assert
image

Braqzen: I see. If we are able to catch the exact error and assert that the specific error matches what we expect in the test then we should do that.

Halimao: > I see. If we are able to catch the exact error and assert that the specific error matches what we expect in the test then we should do that.

Done

Halimao:
@Braqzen @sdankel Hi sir, again for this pr, is there anything that needs to be added?

sdankel (approved):
LGTM, thanks!

PR #526, opened 2023-11-23
feat: support toolchain export

### LINKED ISSUE #322: feat: export a toolchain into `fuel-toolchain.toml`
This is an issue for a proposed new command: `fuelup toolchain export <toolchain>`

This should generate a `fuel-toolchain.toml` in the Sway project root containing

1) toolchain name,
2) components in the toolchain.

Example flow:
```console
$ fuelup default
my-toolchain (default)
$ fuelup show
...
active toolchain
-----------------
my-toolchain (default)
  forc : 0.31.1
    - forc-client
      - forc-deploy : 0.31.1
      - forc-run : 0.31.1
    - forc-doc : 0.31.1
    - forc-explore - not found
    - forc-fmt : 0.31.1
    - forc-index : 0.1.10
    - forc-lsp : 0.31.1
    - forc-wallet - not found
  fuel-core - not found
  fuel-indexer - not found
$ fuelup toolchain export
exported 'my-toolchain' into 'fuel-toolchain.toml'
```

This should then create a TOML file:
```toml
[toolchain]
name = "my-toolchain"

[components]
forc = "0.31.3"  # note that 'forc-fmt', 'forc-lsp', 'forc-doc', 'forc-client' are packaged with 'forc'
forc-index = "0.1.10"
```

The command should fail when there's an existing `fuel-toolchain.toml` in the project root. A `--force` flag should be added to disregard the existing TOML file:

`fuelup toolchain export --force <toolchain>`

Halimao:
close #322 

Support `fuelup toolchain export`

image

image

image

image


src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,98 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    let toolchain_override = ToolchainOverride {
+        cfg: OverrideCfg::new(
+            ToolchainCfg {
+                channel: toolchain_override::Channel::from_str(&toolchain_name).unwrap(),
```
sdankel, line 59: I don't think this will work for custom toolchains. The channel must be one of the distributed channels; it can't be an arbitrary name of a custom toolchain. See `deserialize_channel`

Some options are: 
1. We could default the channel to latest. But, that's making an assumption.
2. We can ask the user to provide a valid channel and then continue the export.
3. We can restrict users to only be able to export their toolchain if they're using a distributed toolchain. This would still be useful for just generating a fuel-toolchain.toml file, as `forc init` currently does not generate this file.
4. In the case where a user supplies `--force` and the file already exists, we could use the `channel` from the previous file, and only override the components where they differ from the user's default toolchain. 

I don't have a strong opinion on this, but I'd lean towards doing 2 & 4.

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,98 @@
...
+use semver::Version;
+use anyhow::{bail, Result};
+use component::{self, Components};
+use tracing::info;
+
+pub fn export(command: ExportCommand) -> Result<()> {
+    let ExportCommand { name, force} = command;
+    let path = PathBuf::from("./").join(FUEL_TOOLCHAIN_TOML_FILE);
+    if !force && path.exists() {
+        bail!("{FUEL_TOOLCHAIN_TOML_FILE} already exists");
```
sdankel, line 19: Let's do a similar behavior to what's being done in https://github.com/FuelLabs/forc-wallet/pull/148

That is, if `--force` is not supplied, ask the user if they wish to override it. If it is, give the user a warning that the file is being overwritten.

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,98 @@
...
+use std::path::{Path, PathBuf};
+use std::str::FromStr;
+use semver::Version;
+use anyhow::{bail, Result};
+use component::{self, Components};
+use tracing::info;
+
+pub fn export(command: ExportCommand) -> Result<()> {
+    let ExportCommand { name, force} = command;
+    let path = PathBuf::from("./").join(FUEL_TOOLCHAIN_TOML_FILE);
```
sdankel, line 17: We should be smarter about locating the toml file. The user should be able to run this command from any level within the forc project.

You can use `get_fuel_toolchain_toml` for this.

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,98 @@
...
+    };
+    let document = toolchain_override.to_toml();
+    std::fs::write(toolchain_override.path, document.to_string())
+        .unwrap_or_else(|_| panic!("failed to write {FUEL_TOOLCHAIN_TOML_FILE}"));
+
+    info!("exported '{toolchain_name}' into '{FUEL_TOOLCHAIN_TOML_FILE}'");
+    Ok(())
+}
+
+fn get_exec_version(component_executable: &Path) -> Result<Version> {
```
sdankel, line 73: How is this different from `exec_version`? Let's move that function to a file called `util/version.rs`, make it public, and import it to use it here.

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,98 @@
...
+        }
+        Err(e) => {
+            if component_executable.exists() {
+                bail!("execute '{} --version' error - {}", component_executable.display(), e);
+            }
+        }
+    }
+
+    bail!("could not show version: {}", component_executable.display());
+}
```
sdankel, line 98: Where are the tests?

sdankel (changes requested):
Thanks for your PR! 🚀

src/commands/toolchain.rs
```diff
@@ -42,6 +46,16 @@ pub struct UninstallCommand {
 #[derive(Debug, Parser)]
 pub struct ListRevisionsCommand {}
 
+#[derive(Debug, Parser)]
+pub struct ExportCommand {
+    /// Toolchain to export, [possible values: latest, beta-1, beta-2, beta-3, beta-4, nightly].
+    /// Your currently active toolchain will be exported if name isn't specified
+    pub name: Option<String>,
+    /// Set true to disregard the existing TOML output file
```
sdankel, line 54:
```suggestion
    /// Forces exporting the toolchain, replacing any existing toolchain override file
```

src/commands/toolchain.rs
```diff
@@ -42,6 +46,16 @@ pub struct UninstallCommand {
 #[derive(Debug, Parser)]
 pub struct ListRevisionsCommand {}
 
+#[derive(Debug, Parser)]
+pub struct ExportCommand {
+    /// Toolchain to export, [possible values: latest, beta-1, beta-2, beta-3, beta-4, nightly].
+    /// Your currently active toolchain will be exported if name isn't specified
```
sdankel, line 52: `Toolchain::from_settings()` gives the default toolchain, not the currently active one.

```suggestion
    /// The default toolchain will be exported if name isn't specified
```

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,267 @@
...
+            "Invalid channel '{}', expected one of <latest-YYYY-MM-DD|nightly-YYYY-MM-DD|beta-1|beta-2|beta-3|beta-4>. \
+            Please input a valid channel: ",
+            toolchain_name,
+        ));
+        let mut input_toolchain_name = String::new();
+        reader.read_line(&mut input_toolchain_name).unwrap();
+        input_toolchain_name = String::from(input_toolchain_name.trim());
+        if toolchain_override::Channel::from_str(&input_toolchain_name).is_err() {
+            bail!(
+                "Invalid channel '{}', expected one of <latest-YYYY-MM-DD|nightly-YYYY-MM-DD|beta-1|beta-2|beta-3|beta-4>.",
```
sdankel, line 96: In the interest of DRY, let's move this to a constant

```
const INVALID_CHANNEL_STR: &str = "Invalid channel '{}', expected one of <latest-YYYY-MM-DD|nightly-YYYY-MM-DD|beta-1|beta-2|beta-3|beta-4>.";
```

sdankel, line 96: This makes me think we should also accept an optional `--channel` argument so the user doesn't have to do this interactively.

sdankel (commented):
Thanks for adding the tests! I left a few more comments, and will test this out next week.

src/ops/fuelup_toolchain/export.rs
Halimao: Done

Halimao: @sdankel Done

Halimao: Note: `toolchain_override::Channel::from_str` logic about validating distributed channels updated, as some invalid channels like "my-custom", "latest-blabla" maybe passed by mistake

image

src/toolchain_override.rs
```diff
@@ -69,7 +69,7 @@ where
         |_| {
             Err(Error::invalid_value(
                 serde::de::Unexpected::Str(&channel_str),
-                &"one of <latest-YYYY-MM-DD|nightly-YYYY-MM-DD|beta-1|beta-2|beta-3|beta-4>",
+                &format!("one of {}", VALID_CHANNEL_STR).as_str(),
```
sdankel, line 72: 👍

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,320 @@
...
+                        };
+                    }
+                }
+            }
+        }
+    }
+    let mut export_channel = toolchain_name.clone();
+    if let Some(channel) = channel {
+        export_channel = channel;
+    }
```
sdankel, line 92:
```suggestion
    let export_channel = channel.unwrap_or(toolchain_name.clone());
```

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,320 @@
...
+                    because a toolchain info file already exists at that location.",
+                    &toolchain_info_path.display(),
+                );
+            }
+        }
+    }
+    let mut toolchain_name = Toolchain::from_settings()?.name;
+    if let Some(name) = name {
+        toolchain_name = name;
+    }
```
sdankel, line 54:
```suggestion
    let toolchain_name = name.unwrap_or(Toolchain::from_settings()?.name);
```

src/commands/toolchain.rs
```diff
@@ -42,6 +50,20 @@ pub struct UninstallCommand {
 #[derive(Debug, Parser)]
 pub struct ListRevisionsCommand {}
 
+#[derive(Debug, Deserialize, Parser)]
+pub struct ExportCommand {
+    /// Toolchain to export, [possible values: latest, beta-1, beta-2, beta-3, beta-4, nightly].
+    /// The default toolchain will be exported if name isn't specified
+    pub name: Option<String>,
+    /// Channel to export, [possible values: latest-YYYY-MM-DD, nightly-YYYY-MM-DD, beta-1, beta-2, beta-3, beta-4].
+    #[clap(value_parser = channel_allowed)]
```
sdankel, line 59: clap will add the possible values to the help message for you:

```suggestion
    /// Channel to export, [possible values: latest-YYYY-MM-DD, nightly-YYYY-MM-DD, beta-1, beta-2, beta-3, beta-4].
    #[clap(value_parser = clap::builder::PossibleValuesParser::new(["latest-YYYY-MM-DD", "nightly-YYYY-MM-DD", "beta-1", "beta-2", "beta-3", "beta-4"]))]
```

src/commands/toolchain.rs
```diff
@@ -64,12 +86,27 @@ fn name_allowed(s: &str) -> Result<String> {
...
+    if toolchain_override::Channel::from_str(s).is_err() {
+        bail!(
+            "Invalid channel '{}', expected one of {}.",
+            s,
+            VALID_CHANNEL_STR,
+        );
+    }
+    Ok(s.to_string())
+}
+
```
sdankel, line 102:
```suggestion
```

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,320 @@
...
+        name,
+        channel,
+        force,
+    } = command;
+    let mut toolchain_info_path = PathBuf::from("./").join(FUEL_TOOLCHAIN_TOML_FILE);
+    if let Some(path) = get_fuel_toolchain_toml() {
+        toolchain_info_path = path;
+        if force {
+            println_warning(&format!(
+                "Because the `--force` argument was supplied, the toolchain info file at {} will be removed.",
```
sdankel, line 28:
```suggestion
                "Because the `--force` argument was supplied, this file will be overwritten: {}",
```

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,320 @@
...
+                &toolchain_info_path.display(),
+            ));
+            fs::remove_file(&toolchain_info_path).unwrap();
+        } else {
+            println_warning(&format!(
+                "There is an existing toolchain info file at {}. \
```
sdankel, line 34:
```suggestion
                "There is an existing toolchain override file at {}. \
```

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,320 @@
...
+                &toolchain_info_path.display(),
+            ));
+            let mut need_replace = String::new();
+            reader.read_line(&mut need_replace).unwrap();
+            if need_replace.trim() == "y" {
+                fs::remove_file(&toolchain_info_path).unwrap();
+            } else {
+                bail!(
+                    "Failed to create a new toolchain info file at {} \
+                    because a toolchain info file already exists at that location.",
```
sdankel, line 45:
```suggestion
                    "Failed to export toolchain \
                    because a toolchain override file already exists at {}.",
```

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,320 @@
...
+        };
+
+        if component.name == component::FORC {
+            let forc_executables = component.executables;
+            for plugin in Components::collect_plugins()? {
+                if !forc_executables.contains(&plugin.name) {
+                    if !plugin.is_main_executable() {
+                        for executable in plugin.executables.iter() {
+                            let plugin_executable = export_toolchain.bin_path.join(executable);
+                            if let Ok(version) = exec_version(plugin_executable.as_path()) {
```
sdankel, line 75: Is `as_path()` needed here?

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,320 @@
...
+        Ok(desc) => Toolchain::from_path(&desc.to_string()),
+        Err(_) => Toolchain::from_path(&toolchain_name),
+    };
+
+    let mut version_map: HashMap<String, Version> = HashMap::new();
+    for component in Components::collect_exclude_plugins()? {
+        let component_executable = export_toolchain.bin_path.join(&component.name);
+        if let Ok(version) = exec_version(component_executable.as_path()) {
+            version_map.insert(component.name.clone(), version);
+        };
```
sdankel, line 66:
```suggestion
        exec_version(export_toolchain.bin_path.join(&component.name))
            .map(|version| version_map.insert(component.name.clone(), version));
```

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,320 @@
...
+                            if let Ok(version) = exec_version(plugin_executable.as_path()) {
+                                version_map.insert(executable.clone(), version);
+                            };
+                        }
+                    } else {
+                        let plugin_executable = export_toolchain.bin_path.join(&plugin.name);
+                        if let Ok(version) = exec_version(plugin_executable.as_path()) {
+                            version_map.insert(plugin.name.clone(), version);
+                        };
+                    }
```
sdankel, line 84: I don't think we don't need to check plugin.is_main_executable() because the behavior is the same.

```suggestion
                    plugin.executables.iter().for_each(|executable| {
                        exec_version(export_toolchain.bin_path.join(executable))
                            .map(|version| version_map.insert(executable.clone(), version));
                    });
```

sdankel (commented):
Looking good, left some nits.

src/commands/toolchain.rs
Halimao: @sdankel  Hi sir, I think using `clap::builder::PossibleValuesParser` would cause unexpected behaviour. Will clap parse "YYYY-MM-DD" correctly? I tried to provide `latest-2023-12-13` but failed.

image

Halimao:
@sdankel Hi sir, looking forward for your review again😀

src/ops/fuelup_toolchain/export.rs
JoshuaBatty, line 1: Thanks @Halimao. Would you mind going back over and removing the unwraps and either handle the error cases, or bail with a message to user?

JoshuaBatty (changes requested):

src/ops/fuelup_toolchain/export.rs
Halimao: > Thanks @Halimao. Would you mind going back over and removing the unwraps and either handle the error cases, or bail with a message to user?

Hi @JoshuaBatty , addressed, thanks for your review~

JoshuaBatty: Thanks, i'll take another look

src/commands/toolchain.rs
```diff
@@ -42,6 +50,20 @@ pub struct UninstallCommand {
 #[derive(Debug, Parser)]
 pub struct ListRevisionsCommand {}
 
+#[derive(Debug, Deserialize, Parser)]
+pub struct ExportCommand {
+    /// Toolchain to export, [possible values: latest, beta-1, beta-2, beta-3, beta-4, nightly].
+    /// The default toolchain will be exported if name isn't specified
+    pub name: Option<String>,
```
Braqzen, line 57: Why does this not allow custom toolchains?

src/commands/toolchain.rs
```diff
@@ -42,6 +50,20 @@ pub struct UninstallCommand {
...
+    /// Channel to export, [possible values: latest-YYYY-MM-DD, nightly-YYYY-MM-DD, beta-1, beta-2, beta-3, beta-4].
```
Braqzen, line 58: Does this also have a default if not specified?

src/commands/toolchain.rs
```diff
@@ -64,12 +86,27 @@ fn name_allowed(s: &str) -> Result<String> {
...
+        return Ok(s.to_string());
+    }
+    if toolchain_override::Channel::from_str(s).is_err() {
+        bail!(
+            "Invalid channel '{}', expected one of {}.",
+            s,
+            VALID_CHANNEL_STR,
+        );
+    }
+    Ok(s.to_string())
```
Braqzen, line 100: Could we return a string slice too?

src/commands/toolchain.rs
```diff
@@ -64,12 +86,27 @@ fn name_allowed(s: &str) -> Result<String> {
     }
 }
 
+fn channel_allowed(s: &str) -> Result<String> {
```
Braqzen, line 89: What is `s`?

Also a bit confused about the returning of an empty `String` version of `s`

src/commands/toolchain.rs
```diff
@@ -64,12 +86,27 @@ fn name_allowed(s: &str) -> Result<String> {
...
+    Ok(s.to_string())
+}
+
 pub fn exec(command: ToolchainCommand) -> Result<()> {
     match command {
         ToolchainCommand::Install(command) => install(command)?,
         ToolchainCommand::New(command) => new(command)?,
         ToolchainCommand::Uninstall(command) => uninstall(command)?,
         ToolchainCommand::ListRevisions(command) => list_revisions(command)?,
+        ToolchainCommand::Export(command) => export(command, stdin().lock())?,
```
Braqzen, line 109: What's the reason for locking the `stdin` here?

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,331 @@
...
+                bail!("failed to remove file {}", &toolchain_info_path.display());
+            }
+        } else {
+            println_warning(&format!(
+                "There is an existing toolchain override file at {}. \
+                Do you wish to replace it with a new one? (y/N) ",
+                &toolchain_info_path.display(),
+            ));
+            let mut need_replace = String::new();
+            if reader.read_line(&mut need_replace).is_err() {
```
Braqzen, line 41: I see. This is probably why we locked above

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,331 @@
...
+            if need_replace.trim() == "y" {
+                if fs::remove_file(&toolchain_info_path).is_err() {
+                    bail!("failed to remove file {}", &toolchain_info_path.display());
+                };
+            } else {
+                bail!(
+                    "Failed to export toolchain \
+                    because a toolchain override file already exists at {}.",
+                    &toolchain_info_path.display(),
+                );
```
Braqzen, line 53: I think a more precise message would be to say that we couldn't remove it because the user has declined the replacement rather than it existed

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,331 @@
...
+            if !fuel_toolchain_toml_file.exists() {
+                fs::create_dir_all(fuel_toolchain_toml_file.parent().unwrap()).unwrap();
+                fs::File::create(&fuel_toolchain_toml_file).unwrap();
+            }
+        } else {
+            fs::File::create(FUEL_TOOLCHAIN_TOML_FILE).unwrap();
+        }
+    }
+    // mock setting.toml
+    fn create_toolchain_settings_file() {
```
Braqzen, line 163: I believe these helpers touch the disk i.e. create and write a file and then remove it.
Does Rust have a library for doing this all in-memory instead? If it does then we could pass the handle around and inspect it instead of reading / writing when testing and potentially leaving behind some files if something exits early.

This is coming from the perspective of having done something like this in Python before

Braqzen (commented):
Question about the exporting format in the images. It looks to be Json-ish.

[Here](https://github.com/FuelLabs/sway-applications/blob/354bac23f11db6b2f89df53b46a8afe43c31911d/AMM/project/fuel-toolchain.toml#L1-L6) is an example of a toolchain file and they look to be formatted differently.

Am I overlooking something or are the formats different when exported?

Halimao:
> Question about the exporting format in the images. It looks to be Json-ish.
> 
> [Here](https://github.com/FuelLabs/sway-applications/blob/354bac23f11db6b2f89df53b46a8afe43c31911d/AMM/project/fuel-toolchain.toml#L1-L6) is an example of a toolchain file and they look to be formatted differently.
> 
> Am I overlooking something or are the formats different when exported?

Exporting format is standard toml format as well, you can check it here: https://www.toml-lint.com/ .
image

src/util/version.rs
```diff
@@ -0,0 +1,30 @@
...
+                None => {
+                    bail!("Error getting version string");
+                }
+            }
+        }
+        Err(e) => {
+            if component_executable.exists() {
+                bail!("execution error - {}", e);
+            } else {
+                bail!("not found");
```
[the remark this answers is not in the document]
Halimao: See screenshot below, just like "Error getting version string", not found will show after component name, meas this component not found. I think it's fine.

image

src/commands/toolchain.rs
Halimao: > Why does this not allow custom toolchains?

Nop, there is no limit for name to be non-custom toolchains, but the channel does

src/commands/toolchain.rs
Halimao: Just as described here "The default toolchain will be exported if name isn't specified" , the default toolchain by `fuelup default` would be used for.

image

src/commands/toolchain.rs
Halimao: Why should this be returned a string slice, do you means that we need to support for export with multi channels?

src/commands/toolchain.rs
Halimao: `channel_allowed` is used as clap `value_parser` for channel, so s means the `channel` option value. And as `--channel` is optional, we should return empty `String`, for unwrap `channel: Option<String>` in `Export`

src/commands/toolchain.rs
Halimao: I have updated to `BufReader::new(stdin())`

src/ops/fuelup_toolchain/export.rs
Halimao: @sdankel What do you think of this, before [suggestion](https://github.com/FuelLabs/fuelup/pull/526#discussion_r1424644440)

src/ops/fuelup_toolchain/export.rs
Halimao: I'm not sure. But `Export`function inner will read file from real disk not the memory.

src/commands/toolchain.rs
Braqzen: Little less work for the compiler if we don't turn it into a String. No biggie

src/commands/toolchain.rs
Braqzen: Would it be more apt to rename to `channel`?

Halimao: I will update it later👌

Halimao: @Braqzen Done

Halimao:
@Braqzen @sdankel Hi sir, is there anything that needs to be added for this pr?

src/ops/fuelup_toolchain/export.rs
sdankel: @Braqzen do you have a suggestion for what the error message should be? It's hard to say that without being too wordy. Maybe something like "Canceled toolchain export." would be enough.

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,331 @@
...
+            }
+        } else {
+            println_warning(&format!(
+                "There is an existing toolchain override file at {}. \
+                Do you wish to replace it with a new one? (y/N) ",
+                &toolchain_info_path.display(),
+            ));
+            let mut need_replace = String::new();
+            if reader.read_line(&mut need_replace).is_err() {
+                bail!("failed to read user input");
```
sdankel, line 42:
```suggestion
                bail!("Failed to read user input");
```

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,331 @@
...
+            }
+            if need_replace.trim() == "y" {
+                if fs::remove_file(&toolchain_info_path).is_err() {
+                    bail!("failed to remove file {}", &toolchain_info_path.display());
```
sdankel, line 46:
```suggestion
                    bail!("Failed to remove file {}", &toolchain_info_path.display());
```

src/ops/fuelup_toolchain/export.rs
sdankel, line 53:
```suggestion
                println_warning("Canceled toolchain export");
                return Ok(());
```

@Braqzen thoughts?

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,331 @@
...
+    let mut export_channel = channel.unwrap_or(toolchain_name.clone());
+    if toolchain_override::Channel::from_str(&export_channel).is_err() {
+        println_warning(&format!(
+            "Invalid channel '{}', expected one of {}. \
+            Please input a valid channel: ",
+            export_channel, VALID_CHANNEL_STR,
+        ));
+        let mut input_channel_name = String::new();
+        if reader.read_line(&mut input_channel_name).is_err() {
+            bail!("failed to read user input");
```
sdankel, line 90: nit
```suggestion
            bail!("Failed to read user input");
```

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,331 @@
...
+                // here shouldn't be err as we has checked above
+                channel: toolchain_override::Channel::from_str(&export_channel)?,
+            },
+            Some(version_map),
+        ),
+        path: toolchain_info_path.clone(),
+    };
+    let document = toolchain_override.to_toml();
+    if std::fs::write(toolchain_override.path, document.to_string()).is_err() {
+        bail!("failed to write {}", toolchain_info_path.display());
```
sdankel, line 116:
```suggestion
        bail!("Failed to write {}", toolchain_info_path.display());
```

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,331 @@
...
+        if toolchain_override
+            .cfg
+            .toolchain
+            .channel
+            .to_string()
+            .eq(channel_name)
+        {
+            Ok(())
+        } else {
+            bail!("unexpected channel");
```
sdankel, line 182:
```suggestion
            bail!("Unknown channel \"{channel_name}\"");
```

src/ops/fuelup_toolchain/export.rs
Braqzen: > Cancelled toolchain export

Is good. It simply states that the action has been discontinued so once the prompt pops up for the user they know that there has been no error. The operation has ended.

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,331 @@
...
+    let mut toolchain_info_path = PathBuf::from("./").join(FUEL_TOOLCHAIN_TOML_FILE);
+    if let Some(path) = get_fuel_toolchain_toml() {
+        toolchain_info_path = path;
+        if force {
+            println_warning(&format!(
+                "Because the `--force` argument was supplied, this file will be overwritten: {}",
+                &toolchain_info_path.display(),
+            ));
+            if fs::remove_file(&toolchain_info_path).is_err() {
+                bail!("failed to remove file {}", &toolchain_info_path.display());
```
sdankel, line 32:
```suggestion
                bail!("Failed to remove file {}", &toolchain_info_path.display());
```

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,331 @@
...
+        ),
+        path: toolchain_info_path.clone(),
+    };
+    let document = toolchain_override.to_toml();
+    if std::fs::write(toolchain_override.path, document.to_string()).is_err() {
+        bail!("failed to write {}", toolchain_info_path.display());
+    }
+
+    info!(
+        "exported '{}' into '{}'",
```
sdankel, line 120:
```suggestion
        "Exported toolchain \"{}\" to file: /"{}/"",
```

src/ops/fuelup_toolchain/export.rs
Braqzen: Probably okay.

I am wondering about general tool design here. Specifically, does it make sense to print a "warning" (maybe it would be better to rename to message) and then to exit with an `Ok(())` if we choose to exit the program early.

In my mind it makes more sense to propagate a specific error all the way up and something would catch that and then handle it by ignoring it or something.

Returning with an `ok` before the intended completion of the workflow is strange to me but the user did choose to cancel so perhaps that can be considered as `ok`.

It's probably too much work to implement this change throughout so it's probably okay as it is although I would generalize to `println_message` unless we have different `println_method`s already

src/ops/fuelup_toolchain/export.rs
sdankel: I think it's okay for the tests to read/write to disk, but the tests should clean up after themselves.

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,331 @@
...
+            .toolchain
+            .channel
+            .to_string()
+            .eq(channel_name)
+        {
+            Ok(())
+        } else {
+            bail!("unexpected channel");
+        }
+    }
```
sdankel, line 184: This is a good use case for [matches](https://doc.rust-lang.org/std/macro.matches.html)

```suggestion
    fn assert_channel_name(expected: &String) {
        let toolchain_override = ToolchainOverride::from_project_root().expect("toolchain override");
        let actual = toolchain_override
            .cfg
            .toolchain
            .channel
            .to_string();
        assert!(matches!(expected, actual));
    }
```

Braqzen (commented):
Running `toolchain export` without any args / options I see the following



I would move the message asking for input onto a new line instead of having a single long line.
The default is not automatically exported, it prompts for input.

The `invalid channel` message comes from a break in backwards compatibility. I haven't followed the changes to know how we manage new toolchains etc. so I'm unsure if this `x86` channel is valid but I have updated fuelup and installed `beta-4` so I assume that it's still valid.
All that to say, we are forcing a new format, have we considered the current / old format and whether we ought to support that?

Running `toolchain export` with `beta-4` I get a file containing the toolchain and components however when I run `latest-YYYY-MM-DD` I get the following error



We ought to handle the difference between a valid channel that the user can type in ex. the beta-X vs the example formats which include a date (as seen above).

According to the possible argument values I should be able to use `latest` however when I do that it tells me that it's an invalid channel.



Overriding an existing toolchain uses the incorrect toolchain. It ignores my input and uses the default now? When I check the content of the file it has `beta-4` so it only prints the incorrect toolchain.

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,319 @@
...
+            println_warning(&format!(
+                "There is an existing toolchain override file at {}. \
+                Do you wish to replace it with a new one? (y/N) ",
+                &toolchain_info_path.display(),
+            ));
+            let mut need_replace = String::new();
+            if reader.read_line(&mut need_replace).is_err() {
+                bail!("Failed to read user input");
+            }
+            if need_replace.trim() == "y" {
```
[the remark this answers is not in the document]
Halimao: > I've recently done a little `stdin` reading. Does this remove spaces and covers new lines? \n and \r?

I think this could be confirmed in the unit test

image

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,319 @@
...
+pub fn export(command: ExportCommand, mut reader: impl BufRead) -> Result<()> {
+    let ExportCommand {
+        name,
+        channel,
+        force,
+    } = command;
+    let mut toolchain_info_path = PathBuf::from("./").join(FUEL_TOOLCHAIN_TOML_FILE);
+    if let Some(path) = get_fuel_toolchain_toml() {
+        toolchain_info_path = path;
+        if force {
```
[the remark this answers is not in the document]
Halimao: Done

src/commands/toolchain.rs
```diff
@@ -1,12 +1,18 @@
-use anyhow::{bail, Result};
+use anyhow::{bail, Ok, Result};
```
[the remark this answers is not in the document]
Halimao: Removed

Halimao:
> Running `toolchain export` without any args / options I see the following
> 
> 
> 
> I would move the message asking for input onto a new line instead of having a single long line. The default is not automatically exported, it prompts for input.
> 
> The `invalid channel` message comes from a break in backwards compatibility. I haven't followed the changes to know how we manage new toolchains etc. so I'm unsure if this `x86` channel is valid but I have updated fuelup and installed `beta-4` so I assume that it's still valid. All that to say, we are forcing a new format, have we considered the current / old format and whether we ought to support that?
> 
> Running `toolchain export` with `beta-4` I get a file containing the toolchain and components however when I run `latest-YYYY-MM-DD` I get the following error
> 
> 
> 
> We ought to handle the difference between a valid channel that the user can type in ex. the beta-X vs the example formats which include a date (as seen above).
> 
> According to the possible argument values I should be able to use `latest` however when I do that it tells me that it's an invalid channel.
> 
> 
> 
> Overriding an existing toolchain uses the incorrect toolchain. It ignores my input and uses the default now? When I check the content of the file it has `beta-4` so it only prints the incorrect toolchain.
> 
> 

I think this should be discussed with @sdankel first. According to the confusion listed above, it seems that you have mixed up the `toolchain name` and `toolchain channel`.

Such as

> Running `toolchain export` without any `args / options` shows the promotions that ask for a valid channel.

Reference to the [comments by @sdankel ](https://github.com/FuelLabs/fuelup/pull/526#discussion_r1412376018)
This is caused by the default toolchain name you have(latest-x86_64_unknown...) is not a valid channel but a valid toolchain name. When there is no --channel provided, we will use the default toolchain name as the channel, and then check the validation of `channel`

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,322 @@
...
+                let mut need_replace = String::new();
+                if reader.read_line(&mut need_replace).is_err() {
+                    bail!("Failed to read user input");
+                }
+                if need_replace.trim() == "y" {
+                    if fs::remove_file(&toolchain_info_path).is_err() {
+                        bail!("Failed to remove file {}", &toolchain_info_path.display());
+                    };
+                } else {
+                    println_warning("Canceled toolchain export");
```
Braqzen, line 51:
```suggestion
                    println_warning("Cancelled toolchain export");
```

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,322 @@
...
+    const INPUT_NOP: &[u8; 1] = b"\n";
+    const INPUT_YES: &[u8; 2] = b"y\r";
+    const INPUT_NO: &[u8; 4] = b"n \r\n";
+    const INPUT_INVALID_CHANNEL: &[u8; 11] = b"my-channel\n";
+    const INVALID_CHANNEL: &str = "my-channel";
+
+    fn remove_toolchain_info() {
+        if let Some(fuel_toolchain_toml_file) = get_fuel_toolchain_toml() {
+            if fuel_toolchain_toml_file.exists() {
+                fs::remove_file(fuel_toolchain_toml_file).unwrap();
```
Braqzen, line 148: Do we want to change the `unwrap()`s to `expect(<msg>)` for clearer failures in tests or are we fine with unwrapping and crashing?

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,322 @@
...
+            fs::create_dir_all(setting_file_path.parent().unwrap()).unwrap();
+            let mut file = fs::File::create(&setting_file_path).unwrap();
+            file.write_all(b"default_toolchain = \"latest\"").unwrap();
+        }
+    }
+    fn assert_channel_name(expected: &String) {
+        let toolchain_override =
+            ToolchainOverride::from_project_root().expect("toolchain override");
+        let _actual = toolchain_override.cfg.toolchain.channel.to_string();
+        assert!(matches!(expected, _actual));
```
Braqzen, line 175:
```suggestion
        let actual = toolchain_override.cfg.toolchain.channel.to_string();
        assert!(matches!(expected, actual));
```

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    #[test]
+    #[serial]
```
Braqzen, line 179: What does this annotation do?

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,322 @@
...
+    fn export_toolchain_with_exists_toolchain_info_with_cancel() {
```
Braqzen, line 180:
```suggestion
    fn export_toolchain_with_existing_toolchain_info_then_cancel() {
```

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,322 @@
...
+            },
+            &INPUT_NO[..],
+        )
+        .unwrap();
+    }
+
+    #[test]
+    #[serial]
+    #[should_panic]
+    fn export_toolchain_with_invalid_channel_provided_throws_err() {
```
Braqzen, line 197:
```suggestion
    fn export_toolchain_with_invalid_channel() {
```

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,322 @@
...
+            },
+            &INPUT_INVALID_CHANNEL[..],
+        )
+        .unwrap();
+    }
+
+    #[test]
+    #[serial]
+    #[should_panic]
+    fn export_toolchain_with_invalid_channel_inputted_throws_err() {
```
Braqzen, line 214:
```suggestion
    fn export_toolchain_without_channel() {
```

src/ops/fuelup_toolchain/export.rs
```diff
@@ -0,0 +1,322 @@
...
+                force: false,
+            },
+            &INPUT_INVALID_CHANNEL[..],
+        )
+        .unwrap();
+    }
+
+    #[test]
+    #[serial]
+    fn export_toolchain_with_exists_toolchain_info() {
```
Braqzen, line 230:
```suggestion
    fn export_toolchain_with_forced_overwrite() {
```

Braqzen: Looking at this and then the following test below I would split them into unit tests because there's more than 1 test case in each.

Braqzen (commented):
I would also split the tests a bit further to group them in order to reduce the cognitive load and make it easier to read.

I would opt for either putting all the success cases at the top and panics at the bottom or consider putting the success cases in a `mod success` and the panics in a `mod panic`.

After the tests are split I'll look over test coverage because I haven't exactly checked if all cases are covered

src/ops/fuelup_toolchain/export.rs
Halimao: reported warning by rustc lint 
> if this is intentional, prefix it with an underscore: `_actual`rustc(unused_variables)

src/ops/fuelup_toolchain/export.rs
Halimao: > What does this annotation do?

For tests that could be guaranteed to be executed in serial, as there are `file creation and deletion`, executing in parallel is non-guaranteed for `file creation and deletion` and causes failed

Halimao:
> I would also split the tests a bit further to group them in order to reduce the cognitive load and make it easier to read.
> 
> I would opt for either putting all the success cases at the top and panics at the bottom or consider putting the success cases in a `mod success` and the panics in a `mod panic`.
> 
> After the tests are split I'll look over test coverage because I haven't exactly checked if all cases are covered

Addressed!

src/ops/fuelup_toolchain/export.rs
sdankel: That's odd, maybe the compiler doesn't recognize the usage inside the `matches` macro. It would be better to use `assert_eq` here anyway.

Halimao: > That's odd, maybe the compiler doesn't recognize the usage inside the `matches` macro. It would be better to use `assert_eq` here anyway.

@sdankel Hi sir, suggestion changes addressed!

sdankel:
This produces a file that looks like this:

```
toolchain = { channel = "beta-3" }
components = { forc = "0.46.1", forc-explore = "0.28.1", forc-tx = "0.46.1", forc-deploy = "0.46.1", forc-wallet = "0.3.0", fuel-core = "0.20.5", forc-run = "0.46.1" }
```

But we want it formatted like this:

```
[toolchain]
channel = "beta-3"
    
[components]
forc-wallet = "0.3.0",
forc-deploy = "0.46.1",
forc-run = "0.46.1",
forc-explore = "0.28.1",
forc-tx = "0.46.1",
forc = "0.46.1",
fuel-core = "0.20.5"
```

sdankel:
I fixed the formatting issue and added tests for it. However I noticed it still pulls the wrong name for the channel by default. When I'm on beta-3, it pulls the full toolchain name rather than the shortened dist name for the channel.

This test should pass:

```
    #[test]
    #[serial]
    fn test_export_beta3_full() {
        create_settings_file();
        create_toolchain_info();

        export(
            ExportCommand {
                name: None,
                channel: Some("beta-3-x86_64-apple-darwin".to_string()),
                force: true,
            },
            &INPUT_NO[..],
        )
        .expect("should succeed");
        assert_toolchain_info(BETA_3_FILE_CONTENTS);
    }
```

Halimao:
> This test should pass:
> 
> ```
>             ExportCommand {
>                 name: None,
>                 channel: Some("beta-3-x86_64-apple-darwin".to_string()),
>                 force: true,
>             },
>             &INPUT_NO[..],
>         )
> ```

One of the valid channel should be `<latest-YYYY-MM-DD|nightly-YYYY-MM-DD|beta-1|beta-2|beta-3|beta-4>`, not sure if this("beta-3-x86_64-apple-darwin" is not a valid channel) is the reason why this test shouldn't pass?

Halimao:
> What I'm saying is if I'm on beta-3, the toolchain name is beta-3-x86_64-apple-darwin. If I just run fuelup toolchain export, I expect it to work. Currently it gives me the error that beta-3-x86_64-apple-darwin is not a valid channel. I think this is because of https://github.com/FuelLabs/fuelup/issues/545

I'm sorry, but this is expected IMO. As we discussed before, 
1. if no toolchain name is provided(run only `fuelup toolchain export`), we will use the default toolchain name, in your local env, that's `beta-3-x86_64-apple-darwin`.
2. Also, as there is no `--channel` provided, we will try to use the toolchain name as channel first. 
3. Then check the channel name, it will prompt user "beta-3-x86_64-apple-darwin is not a valid channel, please input a valid one"
4. If user input `beta-3`, then this will success as `beta-3` is a valid channel, otherwise, abort the process as input channel is invalid

sdankel:
> if no toolchain name is provided(run only fuelup toolchain export), we will use the default toolchain name, in your local env, that's beta-3-x86_64-apple-darwin.

Yes, and we should parse out `beta-3` from the toolchain name using `DistToolchainDescription::from_str` once https://github.com/FuelLabs/fuelup/pull/566 lands.

JoshuaBatty:
This has been implemented in #777

PR #536, opened 2023-12-08
chore: upgrade indexer to v0.24.3

lostman:


ra0x3:
@lostman Conflict btw. (I can merge this since we wanna unblock SwayGang)

Issue #237, opened 2022-10-18
enable specifying a `target` for toolchain installation


bingcicle:
it would be a nice-to-have feature for `fuelup` to install binaries not native to the host machine, even if it's not an immediate requirement.

PR #566, opened 2024-02-02
Improve DistToolchainDescription

### LINKED ISSUE #545: Fuelup fails to parse beta toolchains
[This function](https://github.com/FuelLabs/fuelup/blob/96d7fb248e436bedb467fcc8197e2039b44aabf4/src/toolchain.rs#L117) seems to be totally broken. For example, it fails to parse any of the beta toolchain names:

```
    #[test]
    fn test_parse_name_apple() {
        let desc = DistToolchainDescription::from_str("beta-3-x86_64-apple-darwin").unwrap();
        assert_eq!(desc.name, DistToolchainName::from_str("beta-3").unwrap());
        assert_eq!(desc.date, None);
        assert_eq!(desc.target, Some(TargetTriple::new("x86_64-apple-darwin").unwrap()));
    }
```

Unless I'm reading this wrong, it seems like this would cause a lot of usability issues.

crodas:
Fixes #545 

Improved the parsing logic to be resilient to names with dashes in it.

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -29,6 +29,7 @@ pub const RESERVED_TOOLCHAIN_NAMES: &[&str] = &[
     channel::BETA_4,
     channel::BETA_5,
     channel::NIGHTLY,
+    // Stable is reserved, although currently unused.
     channel::STABLE,
 ];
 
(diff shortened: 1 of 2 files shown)


src/toolchain.rs
```diff
@@ -116,39 +116,88 @@ impl fmt::Display for DistToolchainDescription {
...
+                },
+                Err(_) => TargetTriple::from_host().ok(),
+            }
+        }
+    }
+}
+
 impl FromStr for DistToolchainDescription {
     type Err = anyhow::Error;
     fn from_str(s: &str) -> Result<Self> {
```
sdankel, line 174: It would be good to add a comment about what formats of `s` are supported, since this is confusing

sdankel:
Closes https://github.com/FuelLabs/fuelup/issues/237

src/toolchain.rs
sdankel: To save time going back and forth, I updated the comment and tests to be more comprehensive: https://github.com/FuelLabs/sway/pull/5544

They're all passing so I'm happy to approve once those changes are merged in 😃

sdankel (approved):

PR #568, opened 2024-02-07
add html redirect to install script

JoshuaBatty:
This PR introduces an index.html file at the root of the gh-pages branch to facilitate automatic redirection to the `fuelup-init.sh` script when the root URL (https://install.fuel.network/) is accessed. This change simplifies the process for users to download and install `fuelup` by allowing a direct curl command from the root URL.

The motivation is to change the install command from 

```
curl -sSf https://install.fuel.network/fuelup-init.sh | sh
```

to this

```
curl -sSf https://fuel.network/install | sh
```

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <title>Redirecting...</title>
+  <script>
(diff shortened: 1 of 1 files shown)


sdankel (approved):

sdankel:
utACK

PR #573, opened 2024-02-19
Print warning if there is no toolchain to update

### LINKED ISSUE #560: fuelup update

After uninstalling all toolchains running the update command prints empty lines instead of informing us that there are no toolchains to update

### Solution you'd like

image

### Notes

_No response_

crodas:
Fixes #560

### DIFF
--- a/src/ops/fuelup_update.rs
+++ b/src/ops/fuelup_update.rs
@@ -1,14 +1,14 @@
 use crate::{
     channel::Channel,
     config::Config,
-    fmt::{bold, colored_bold},
+    fmt::{bold, colored_bold, println_error},
     path::warn_existing_fuel_executables,
     toolchain::{DistToolchainDescription, Toolchain},
 };
 use ansiterm::Color;
 use anyhow::{bail, Result};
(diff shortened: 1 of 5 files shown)


src/ops/fuelup_update.rs
```diff
@@ -20,6 +20,13 @@ pub fn update() -> Result<()> {
 
     warn_existing_fuel_executables()?;
 
+    if toolchains.is_empty() {
+        error!(
+            "Could find any toolchain. Please run `fuelup default <toolchain>` to install toolchains first."
```
sdankel, line 25:
```suggestion
        println_error(format!(
            "No toolchains are installed. Use `fuelup default <toolchain>` to install a toolchain."
```

src/ops/fuelup_update.rs
```diff
@@ -20,6 +20,13 @@ pub fn update() -> Result<()> {
 
     warn_existing_fuel_executables()?;
```
sdankel, line 21: This function should also use `println_warning`

sdankel (approved):

PR #574, opened 2024-02-19
Remove files from store when uninstalling

### LINKED ISSUE #559: fuelup toolchain uninstall <name>

Similar comments to the default command

image

### Solution you'd like

When uninstalling the last available toolchain there is a formatting error (a tab seems to be used).

Instead of telling the user that we have failed to set a new default toolchain, inform them that there are no more toolchains and they should create another one. Moreover, the last sentence does not make sense in this context because there are no more toolchains so they cannot manually set one.

image

Uninstalling a toolchain only deletes the directory and the copied binaries it contains. It does not delete those components from the `store`.

If a user uninstalls an old toolchain they likely want to remove all remnants so those files should be removed from the `store`. At the same time, they may have multiple toolchains which mix versions so we do not want to delete the components from `store` until we check all toolchains an no toolchain uses that component

### Notes

_No response_

crodas:
Fixes #559

### DIFF
--- a/tests/toolchain.rs
+++ b/tests/toolchain.rs
@@ -126,8 +126,10 @@ fn fuelup_toolchain_uninstall() -> Result<()> {
 
         for toolchain in &toolchains[1..2] {
             let toolchain_with_target = format_toolchain_with_target(toolchain);
+            assert!(cfg.toolchains_dir().join(&toolchain_with_target).is_dir());
             let output = cfg.fuelup(&["toolchain", "uninstall", toolchain]);
             let expected_stdout = format!("Toolchain '{toolchain_with_target}' uninstalled\n");
+            assert!(!cfg.toolchains_dir().join(toolchain_with_target).is_dir());
             assert!(
                 output.stdout.contains(&expected_stdout),
                 "toolchain: {}",


src/ops/fuelup_toolchain/uninstall.rs
```diff
@@ -36,8 +36,9 @@ pub fn uninstall(command: UninstallCommand) -> Result<()> {
                 }
 
                 error!(
-                "Could not set default toolchain after uninstallation of currently used toolchain. 
-                Please run `fuelup default <toolchain>` to manually switch your current toolchain."
+                    "{}\r\t{}",
+                    "Could not set default toolchain after uninstallation of currently used toolchain",
```
sdankel, line 40: It seems like we shouldn't let the user uninstall their current default toolchain. What do you think?

JoshuaBatty: Yeah I agree with Sophie. There should always be at least one toolchain installed and that should be the default.

src/ops/fuelup_toolchain/uninstall.rs
```diff
@@ -20,13 +20,17 @@ pub fn uninstall(command: UninstallCommand) -> Result<()> {
     };
 
     if !toolchain.exists() {
-        info!("toolchain '{}' does not exist", &toolchain.name);
+        println_warn(format!("Toolchain '{}' does not exist", &toolchain.name));
         return Ok(());
     }
 
+    if config.list_toolchains()?.len() == 1 {
+        bail!("Cannot uninstall the last toolchain");
```
sdankel, line 28:
```suggestion
        bail!("Cannot uninstall {}  as it is currently the default toolchain. Run `fuelup default <toolchain>` to update the default toolchain.", &toolchain.name);
```

src/ops/fuelup_toolchain/uninstall.rs
```diff
@@ -20,13 +20,17 @@ pub fn uninstall(command: UninstallCommand) -> Result<()> {
...
+    }
+
     match toolchain.uninstall_self() {
         Ok(_) => {
-            info!("toolchain '{}' uninstalled", &toolchain.name);
+            println!("Toolchain '{}' uninstalled", &toolchain.name);
```
sdankel, line 33:
```suggestion
            info!("Toolchain '{}' uninstalled", &toolchain.name);
```

We shouldn't always use tracing instead of printlns so that all output is captured in the log files.

src/ops/fuelup_toolchain/uninstall.rs
```diff
@@ -20,13 +20,17 @@ pub fn uninstall(command: UninstallCommand) -> Result<()> {
     };
 
     if !toolchain.exists() {
-        info!("toolchain '{}' does not exist", &toolchain.name);
+        println_warn(format!("Toolchain '{}' does not exist", &toolchain.name));
         return Ok(());
     }
 
+    if config.list_toolchains()?.len() == 1 {
```
sdankel, line 27: From this comment: https://github.com/FuelLabs/fuelup/pull/574/files#r1495080029

This should be something like:

```
    if toolchain.name == settings.default_toolchain {
```

src/toolchain.rs
```diff
@@ -431,6 +431,19 @@ impl Toolchain {
...
+            .into_iter()
+            .filter(|component| self.has_component(&component.name))
+            .filter_map(|component| {
+                get_bin_version(&self.bin_path.join(&component.name))
+                    .ok()
+                    .map(|version| store.component_dir_path(&component.name, &version))
+            })
+            .for_each(|p| {
+                let _ = remove_dir_all(p);
+            });
```
sdankel, line 445: This would be cleaner as a function of `Store`'s implementation so we can do `store.remove_component(&component.name, &version)`

src/ops/fuelup_toolchain/uninstall.rs
JoshuaBatty: Instead of making sure there is only one toolchain left, it should always be the default toolchain.

tests/toolchain.rs
```diff
@@ -111,33 +111,34 @@ fn fuelup_toolchain_uninstall() -> Result<()> {
...
+        let expected_stdout = "as it is currently the default toolchain. Run `fuelup default <toolchain>` to update the default toolchain.";
+        assert!(output.stdout.contains(expected_stdout));
+
+        for toolchain in &toolchains[1..2] {
             let toolchain_with_target = format_toolchain_with_target(toolchain);
             let output = cfg.fuelup(&["toolchain", "uninstall", toolchain]);
-            let expected_stdout = format!("toolchain '{toolchain_with_target}' uninstalled\n");
-
-            assert!(output.stdout.contains(&expected_stdout));
+            let expected_stdout = format!("Toolchain '{toolchain_with_target}' uninstalled\n");
```
sdankel, line 130: nit: It would also be good to assert that the correct directories were removed.

sdankel (commented):
This is looking good! Just one small comment about the test, then happy to approve.

JoshuaBatty (approved):

PR #575, opened 2024-02-19
Enhance to toolchain install

### LINKED ISSUE #558: fuelup toolchain install <name>

When you install a new toolchain, such as beta-3, for the first time it downloads the components as expected.

When you re-run the command right after (to check for “updates”, as the doc string suggests) it tells the user that components are re-added but no download logs are shows (Fetching <component> from <url>).

Instead, check the versions and remove all but the last line which informs me that the components are already up to date.

image

Note: installing a new toolchain does not automatically set it as the default. I assumed that it would, user preference here.

### Solution you'd like

read above

### Notes

_No response_

crodas:
Fixes #558

When updating the current toolchain this used to be output

image

This is the output now:

image

This change only affects updates, new installation have the same output

image

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -98,7 +98,8 @@ fn consume_back<T>(parts: &mut VecDeque<T>, number: usize) {
     }
 }
 
-/// Attempts to parse a date from the front of the parts list, returning the date and consuming the date parts if they are available
+/// Attempts to parse a date from the front of the parts list, returning the date and consuming the
+/// date parts if they are available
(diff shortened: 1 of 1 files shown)


sdankel (approved):
Nice! 

Another enhancement would be to only show "Downloading: {}" and the download bars, none of the other messages, unless the `verbose` flag is given. I did a similar thing in forc with the tracing library, so that we only need to change the `info!`s to `debug!`s for the things we want to hide.

PR #576, opened 2024-02-19
Do not download self update if it is the same version

### LINKED ISSUE #555: `fuelup self update` downloading the same version

Despite having the same version already installed we appear to download the same release again.

image

### Solution you'd like

This may be an intentional decision to “fix” the tool by reinstalling if something is broken, if not then add a check on the version before downloading again to save on bandwidth and work done on the filesystem

### Notes

_No response_

crodas:
Fixes #555

After:

image

### DIFF
--- a/src/ops/fuelup_check.rs
+++ b/src/ops/fuelup_check.rs
@@ -56,26 +56,12 @@ fn format_version_comparison(current_version: &Version, latest_version: &Version
     }
 }
 
-fn check_plugin(plugin_executable: &Path, plugin: &str, latest_version: &Version) -> Result<()> {
-    match get_bin_version(plugin_executable) {
-        Ok(version) => {
-            info!(
-                "{:>4}- {} - {}",
-                "",
-                bold(plugin),
-                format_version_comparison(&version, latest_version)
-            );
-        }
-        Err(err) => {
-            info!(
-                "{:>4}- {} - Error getting version string: {}",
-                "",
-                plugin,
-                err.to_string()
-            );
-        }
-    }
-    Ok(())
+fn check_plugin(plugin_executable: &Path, plugin: &str, latest_version: &Version) {
+    let version_or_err = match get_bin_version(plugin_executable) {
+        Ok(version) => format_version_comparison(&version, latest_version),
+        Err(err) => err.to_string(),
+    };
+    info!("{:>4}- {} - {}", "", plugin, version_or_err);
 }
 
 fn check_fuelup() -> Result<()> {
@@ -110,24 +96,11 @@ fn check_toolchain(toolchain: &str, verbose: bool) -> Result<()> {
     for component in Components::collect_exclude_plugins()? {
         if let Some(latest_version) = latest_package_versions.get(&component.name) {
             let component_executable = toolchain.bin_path.join(&component.name);
-            match get_bin_version(&component_executable) {
-                Ok(version) => {
-                    info!(
-                        "{:>2}{} - {}",
-                        "",
-                        bold(&component.name),
-                        format_version_comparison(&version, latest_version)
-                    );
-                }
-                Err(e) => {
-                    error!(
-                        "{:>2}{} - Error getting version string: {}",
-                        "",
-                        bold(&component.name),
-                        e.to_string()
-                    );
-                }
-            }
+            let version_text = match get_bin_version(&component_executable) {
+                Ok(version) => format_version_comparison(&version, latest_version),
+                Err(err) => err.to_string(),
+            };
+            info!("{:>2}{} - {}", "", bold(&component.name), version_text);
             if verbose && component.name == component::FORC {
                 for plugin in component::Components::collect_plugins()? {
                     if !plugin.is_main_executable() {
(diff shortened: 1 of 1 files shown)


src/file.rs
```diff
@@ -9,6 +10,21 @@ pub(crate) fn is_executable(file: &Path) -> bool {
     file.is_file() && file.metadata().unwrap().permissions().mode() & 0o111 != 0
 }
 
+pub(crate) fn get_bin_version(exec_path: &Path) -> Option<Version> {
```
sdankel, line 13: 💟

src/ops/fuelup_self.rs
```diff
@@ -27,6 +27,14 @@ pub fn self_update() -> Result<()> {
 
     let fuelup_bin = fuelup_bin();
 
+    if !force && get_bin_version(&fuelup_bin) == Some(download_cfg.version.clone()) {
+        info!(
+            "fuelup is already up to date at version {}",
```
sdankel, line 32: nit:
```suggestion
            "Already up to date (fuelup v{})",
```

src/ops/fuelup_show.rs
```diff
@@ -1,11 +1,12 @@
-use anyhow::{bail, Result};
+use anyhow::Result;
 use component::{self, Components};
 use semver::Version;
 use std::collections::HashMap;
-use std::path::Path;
+
```
sdankel, line 5:
```suggestion
```

src/ops/fuelup_show.rs
```diff
@@ -1,11 +1,12 @@
...
 use std::str::FromStr;
 use tracing::info;
```
sdankel, line 8:
```suggestion
```

src/ops/fuelup_show.rs
```diff
@@ -17,33 +18,6 @@ use crate::{
...
-                    Ok(version)
-                }
-                None => {
-                    bail!("Error getting version string");
-                }
-            }
-        }
-        Err(e) => {
-            if component_executable.exists() {
-                bail!("execution error - {}", e);
```
sdankel, line 39: This is still a useful error. Could we incorporate this error handling into `get_bin_version` and have it return a Result instead of Option?

src/toolchain.rs
```diff
@@ -304,7 +308,7 @@ impl Toolchain {
         let fuelup_bin = fuelup_bin();
         if !fuelup_bin.is_file() {
             info!("fuelup not found - attempting to self update");
```
sdankel, line 310: This doesn't seem like it should be possible 🤣 If someone wanted to install it only from cargo, this would not update the cargo version, and install fuelup in the home directory, which is probably not what they want. 

In that case I think we should give an error message like `fuelup cannot update itself because it was installed by a package manager. For full functionality, visit https://docs.fuel.network/docs/fuelup/installation and re-install.`

tests/self.rs
```diff
@@ -5,12 +5,22 @@ use testcfg::FuelupState;
...
+        let output = cfg.fuelup(&["self", "update", "--force"]);
         let expected_stdout_starts_with = "Fetching binary from";
         assert!(output.stdout.starts_with(expected_stdout_starts_with));
     })?;
 
     Ok(())
 }
+
+#[test]
+fn fuelup_self_update_latest() -> Result<()> {
```
sdankel, line 18: 👍

sdankel (commented):
Mostly nits, thanks for doing this!

src/file.rs
```diff
@@ -1,14 +1,38 @@
...
+use semver::Version;
+use std::{fs, io, os::unix::fs::PermissionsExt, path::Path};
 
 #[cfg(unix)]
 pub(crate) fn is_executable(file: &Path) -> bool {
     file.is_file() && file.metadata().unwrap().permissions().mode() & 0o111 != 0
 }
 
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum BinError {
```
sdankel, line 11: This is much better!

src/ops/fuelup_check.rs
```diff
@@ -56,34 +57,22 @@ fn format_version_comparison(current_version: &Version, latest_version: &Version
...
-                        "",
-                        bold(plugin),
-                        format_version_comparison(&version, latest_version)
-                    );
-                }
-                None => {
-                    info!("{:>4}- {} - Error getting version string", "", plugin);
-                }
-            };
+    match get_bin_version(plugin_executable) {
```
sdankel, line 60: nit: we can make it even shorter! This function never returns an error, so doesn't need to return an empty result.

```rust
fn check_plugin(plugin_executable: &Path, plugin: &str, latest_version: &Version) {
    let version_or_err = match get_bin_version(plugin_executable) {
        Ok(version) => format_version_comparison(&version, latest_version),
        Err(err) => err.to_string()
    };
    info!("{:>4}- {} - {}", "", plugin, version_or_err);
}
```

src/ops/fuelup_check.rs
```diff
@@ -121,36 +110,24 @@ fn check_toolchain(toolchain: &str, verbose: bool) -> Result<()> {
...
-
+                Err(e) => {
+                    error!(
+                        "{:>2}{} - Error getting version string: {}",
+                        "",
+                        bold(&component.name),
+                        e.to_string()
+                    );
+                }
+            }
```
sdankel, line 130: Same thing here:

```rust 
            let version_text = match get_bin_version(&component_executable) {
                Ok(version) => format_version_comparison(&version, latest_version),
                Err(err) => err.to_string(),
            };
            info!("{:>2}{} - {}", "", bold(&component.name), version_text);
```

sdankel (commented):
Looking good, just a couple nits 👍

sdankel (approved):

PR #585, opened 2024-02-29
Add support for self uninstall

### LINKED ISSUE #500: Add a `fuelup self uninstall` command to uninstall `fuelup`
It would be nice if we had an easy way to install `fuelup` just like `rustup` has. See below for how rust does this.

```
% rustup self uninstall
Thanks for hacking in Rust!

This will uninstall all Rust toolchains and data, and remove
$HOME/.cargo/bin from your PATH environment variable.

Continue? (y/N) y

info: removing rustup home
info: removing cargo home
```

crodas:
Fixes #500

Adds support for `fuelup self uninstall`

### DIFF
--- a/src/ops/fuelup_self.rs
+++ b/src/ops/fuelup_self.rs
@@ -23,25 +23,45 @@ pub fn attempt_install_self(download_cfg: DownloadCfg, dst: &Path) -> Result<()>
     Ok(())
 }
 
+#[inline]
+fn remove_path_from_content(file_content: &str) -> (bool, String) {
+    let mut is_modified = false;
+    let whole_definition = format!("PATH={}", FUELUP_DIR);
+    let suffix = format!("{}:", FUELUP_DIR);
+    let prefix = format!("{}:", FUELUP_DIR);
+    let lines = file_content
+        .trim_end_matches('\n')
+        .trim_end_matches('\r')
+        .lines()
+        .collect::<Vec<_>>();
+    let total_lines = lines.len();
+    let new_lines = lines
+        .into_iter()
+        .filter(|line| !line.trim_end().ends_with(&whole_definition))
+        .map(|line| {
+            if line.contains("PATH") && line.contains(FUELUP_DIR) {
+                is_modified = true;
+                line.trim()
+                    .replace(&suffix, "")
+                    .replace(&prefix, "")
+                    .replace(FUELUP_DIR, "")
+            } else {
+                line.to_owned()
+            }
+        })
+        .collect::<Vec<_>>();
+    (
+        is_modified || total_lines != new_lines.len(),
+        new_lines.join("\n"),
+    )
+}
+
 /// Removes the fuelup directory from $PATH
 fn remove_fuelup_from_path() -> Result<()> {
     for shell in Shell::all() {
         for rc in shell.rc_files().into_iter().filter(|c| c.is_file()) {
-            let file = read_file("rcfile", &rc)?;
-            let mut is_modified = false;
-            let new_content = file
-                .lines()
-                .filter(|line| {
-                    if line.contains("PATH") && line.contains(FUELUP_DIR) {
-                        is_modified = true;
-                        false
-                    } else {
-                        true
-                    }
-                })
-                .collect::<Vec<_>>()
-                .join("\n");
-            if is_modified {
+            let (was_modified, new_content) = remove_path_from_content(&read_file("rcfile", &rc)?);
+            if was_modified {
                 println_info(format!(
                     "{} has been updated to remove fuelup from $PATH",
                     rc.display()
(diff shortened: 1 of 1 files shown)


src/commands/fuelup.rs
```diff
@@ -15,10 +23,18 @@ pub struct UpdateCommand {
...
     if let Err(e) = self_update(force) {
         bail!("fuelup failed to update itself: {}", e)
     };
 
     Ok(())
 }
+
+pub fn remove_exec(force: bool) -> Result<()> {
+    if let Err(e) = self_uninstall(force) {
+        bail!("fuelup failed to update itself: {}", e)
```
sdankel, line 36:
```suggestion
        bail!("Failed to uninstall fuelup: {}", e)
```

src/commands/fuelup.rs
```diff
@@ -15,10 +23,18 @@ pub struct UpdateCommand {
     pub force: bool,
 }
 
-pub fn exec(force: bool) -> Result<()> {
+pub fn update_exec(force: bool) -> Result<()> {
     if let Err(e) = self_update(force) {
         bail!("fuelup failed to update itself: {}", e)
```
sdankel, line 28:
```suggestion
        bail!("Failed to update fuelup: {}", e)
```

src/ops/fuelup_self.rs
```diff
@@ -18,6 +22,38 @@ pub fn attempt_install_self(download_cfg: DownloadCfg, dst: &Path) -> Result<()>
...
 
+const GOODBYE_HEADER: &str = "Thanks for hacking in Sway!";
+const GOODBYE_INFO: &str = "This will uninstall all Sway toolchains and data, and remove, $HOME/.fuelup/bin from your PATH environment variable.";
+
+pub fn self_uninstall(force: bool) -> Result<()> {
+    println!("{}\n\n{}", GOODBYE_HEADER, GOODBYE_INFO);
+    if force || ask_user_yes_no_question("Continue? (y/N)").context("Console I/O")? {
+        let remove = [
+            ("remove forc home", forc_dir()),
+            ("remove fuel home", fuel_dir()),
```
sdankel, line 33:
```suggestion
            ("removing forc home", forc_dir()),
            ("removing fuel home", fuel_dir()),
```

src/ops/fuelup_self.rs
```diff
@@ -18,6 +22,38 @@ pub fn attempt_install_self(download_cfg: DownloadCfg, dst: &Path) -> Result<()>
     Ok(())
 }
 
+const GOODBYE_HEADER: &str = "Thanks for hacking in Sway!";
+const GOODBYE_INFO: &str = "This will uninstall all Sway toolchains and data, and remove, $HOME/.fuelup/bin from your PATH environment variable.";
```
sdankel, line 26: I don't see where we're modifying the PATH. We should give users a separate choice about modifying the PATH if they don't use `--force`

src/path.rs
```diff
@@ -9,12 +9,22 @@ use std::{
 };
 
 pub const FUELUP_DIR: &str = ".fuelup";
+pub const FORC_HOME: &str = ".forc";
+pub const FUEL_HOME: &str = ".fuel";
```
sdankel, line 13: When does this get created?

src/ops/fuelup_self.rs
sdankel, line 26:
```suggestion
const GOODBYE_INFO: &str = "This will uninstall all Sway toolchains and data, and optionally remove $HOME/.fuelup/bin from your PATH environment variable.";
```

src/ops/fuelup_self.rs
```diff
@@ -18,6 +22,38 @@ pub fn attempt_install_self(download_cfg: DownloadCfg, dst: &Path) -> Result<()>
...
+
+pub fn self_uninstall(force: bool) -> Result<()> {
```
sdankel, line 28: We should add unit tests for this.

src/path.rs
JoshuaBatty: I don't think we should be removing these folders only the fuelup one. 

If I have installed forc from source then I would still want these folders present.

src/ops/fuelup_self.rs
sdankel: What do you mean? If we tell users that PATH is going to be modified, we should actually do it.

JoshuaBatty: Thanks @crodas Are we able to write a test for the code added in https://github.com/FuelLabs/fuelup/commit/d8c49245f3d5117fd41028e2a0755661783cce20

src/ops/fuelup_self.rs
```diff
@@ -18,6 +23,68 @@ pub fn attempt_install_self(download_cfg: DownloadCfg, dst: &Path) -> Result<()>
...
+    );
+    if force || ask_user_yes_no_question("Continue? (y/N)").context("Console I/O")? {
+        let remove = [
+            ("removing fuelup binaries", fuelup_bin_dir()),
+            ("removing fuelup home", fuelup_dir()),
+        ];
+        remove_fuelup_from_path()?;
+
+        for (info, path) in remove.into_iter() {
+            println_warn(info);
```
JoshuaBatty, line 70: Can we change this to `info` rather than warning so it matches rustup.

Currently: fuelup

```
warning: /Users/joshuabatty/.zshrc has been updated to remove fuelup from $PATH
warning: removing fuelup binaries
warning: removing fuelup home
```

rustup

```
info: removing rustup home
info: removing cargo home
info: removing rustup binaries
info: rustup is uninstalled
```

JoshuaBatty (changes requested):

PR #631, opened 2024-06-18
Alter use of the --verbose flag for `fuelup check`

### LINKED ISSUE #553: fuelup check —verbose

The current message makes me wonder if the plugins will be displayed but without their versions
image

### Solution you'd like

Rephrase the `--verbose` description to `Include forc plugin versions alongside forc` or similar.

### Notes

_No response_

JoshuaBatty:
The output of `fuelup check --verbose` is what I would expect `fuelup check` to output. Before this change, `fuelup check` outputted a fraction of the installed components on the system. ~~I feel like this just leads to confusion amongst users so it's been removed.~~

I added the -v flag back in. The default behaviour is to only show fuelup or components need updating. The -v flag shows all components and prints that they are up to date. See screen shot below. 

The rest of the diff is just cleaning up imports and doing clippy. 

Screenshot 2024-06-25 at 1 59 44 pm

closes #553


sdankel:
I agree that we shouldn't need a verbose flag for `fuelup check`, but I also think we should condense its output. I think it should show 1 line per installed toolchain, rather than showing every plugin in the toolchain, similar to rustup.

image

image

Currently it prints a download bar even when everything is up to date. On that note, `check` should never download anything, it should only tell the user if something can be updated, so that's a bug.

JoshuaBatty:
Thanks Sophie. I like the idea of mimicking what `cargo check` does here. I'll also look into avoiding need to download anything for this check. Putting into draft mode until this is finished.

JoshuaBatty:
Thanks @sdankel this should be ready now. I added the -v flag back in so it prints if the component of fuelup is up to date, but the default behaviour is now similar to rustup and only shows if something is out of date. 

I also removed the download bar. We still need to download the toml file from git to check the versions but it's only a handful of kb so showing a download bar seems unnecessary.

src/ops/fuelup_check.rs
```diff
@@ -53,95 +49,94 @@ fn format_version_comparison(current_version: &Version, latest_version: &Version
...
 }
 
 fn check_toolchain(toolchain: &str, verbose: bool) -> Result<()> {
     let description = DistToolchainDescription::from_str(toolchain)?;
-
     let dist_channel = Channel::from_dist_channel(&description)?;
     let latest_package_versions = collect_package_versions(dist_channel);
-
     let toolchain = Toolchain::new(toolchain)?;
+    info!("{}: {}", bold("Toolchain: "), &toolchain.name);
```
sdankel, line 93:
```suggestion
    info!("{}{}", bold("Toolchain: "), &toolchain.name);
```

image

I get the need for this label, because with all of the components listed below it it's hard to see that that's the toolchain name. It's a lot more readable without the download bar! 👍 

Since we have the `-v` added back, I wonder if we should go further with condensing the output in non-verbose mode. It could be like this:

```
latest-aarch64-apple-darwin - Up to date
nightly-aarch64-apple-darwin - Update available (12 updates)
```

sdankel (commented):
Looking much better! One thing I noticed is when I have 1 toolchain that needs updates and 1 that is up to date, the formatting for the up-to-date one is off, and doesn't say up-to-date:

image

src/ops/fuelup_check.rs
JoshuaBatty: I like this idea :)

JoshuaBatty: putting into draft mode until I complete this

JoshuaBatty: done in [05a3363](https://github.com/FuelLabs/fuelup/pull/631/commits/05a336343b91089d0b0156d0567c99053c284121)

JoshuaBatty:
Thanks @sdankel i've update the description with a screenshot of what the default and verbose options produce.

fuel-service-user:
### [LCOV](https://github.com/marketplace/actions/report-lcov) of commit [<code>9bde78c</code>](631/commits/9bde78c5bb31b65ecfa46c0d6b7cd1dd44303cf6) during [CI #2025](../actions/runs/9669662618)
<pre>Summary coverage rate:
  lines......: 81.9% (2315 of 2827 lines)
  functions..: 44.4% (363 of 818 functions)
  branches...: 58.6% (278 of 474 branches)

sdankel (approved):
LGTM 🚀

PR #632, opened 2024-06-18
Show installed toolchains if a user tries to set an incorrect toolchain name to the default.

### LINKED ISSUE #557: fuelup default <name>

When choosing a name that does not exist either prompt the user to use the `fuelup show` command to check the toolchains or preferably automatically show the installed toolchains below.

image
image
Capitalize the `d` in `default`

image

### Solution you'd like

When choosing a name that does not exist either prompt the user to use the `fuelup show` command to check the toolchains or preferably automatically show the installed toolchains below.

### Notes

_No response_

JoshuaBatty:
closes #557


tests/check.rs
```diff
@@ -27,6 +27,7 @@ fn fuelup_check() -> Result<()> {
         let output = cfg.fuelup(&["check"]);
         let stripped = strip_ansi_escapes::strip(output.stdout);
         let stdout = String::from_utf8_lossy(&stripped);
+        println!("\n{}", &stdout);
```
sdankel, line 30:
```suggestion
```

sdankel (approved):

tests/check.rs
JoshuaBatty: Good catch :)

sdankel (approved):

Issue #634, opened 2024-06-19
Panic in fuelup "no entry found for key"


sdankel:
![image](https://github.com/FuelLabs/fuelup/assets/47993817/6dacf499-086d-44a0-8d79-e1f17e0eaaaa)

PR #635, opened 2024-06-19
fix: Fix panic in fuelup download

sdankel:
Closes https://github.com/FuelLabs/fuelup/issues/634

The problem is that BTreeMap index accession with `[]` doesn't return a result or option, it just panics if it's not there.

It would be good to have a lint to require us to use `get()` instead of index access, to avoid panics like this: https://github.com/FuelLabs/fuelup/issues/636

### DIFF
--- a/src/download.rs
+++ b/src/download.rs
@@ -1,25 +1,27 @@
+use crate::{
+    channel::{Channel, Package},
+    constants::CHANNEL_LATEST_URL,
+    target_triple::TargetTriple,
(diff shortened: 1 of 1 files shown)


kayagokalp (approved):
Realized that nightly macos x86 builds are failing and missing from nightly entries https://github.com/FuelLabs/fuelup/blob/gh-pages/channel-fuel-nightly.toml, so anyone using nightly channel from x86 mac os should have an issue but this will be makin sure that we have a nice error!

PR #639, opened 2024-06-25
Add the indoc crate to help clean up examples

### LINKED ISSUE #638: Add the `indoc` crate to help clean up examples
[https://crates.io/crates/indoc](https://crates.io/crates/indoc)

JoshuaBatty:
closes #638

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -326,6 +326,7 @@ name = "component"
 version = "0.0.0"
 dependencies = [
  "anyhow",
+ "indoc",
  "serde",
  "toml_edit 0.13.4",
 ]
@@ -505,6 +506,7 @@ dependencies = [
  "dirs",
  "flate2",
  "indicatif",
+ "indoc",
  "semver",
  "serde",
  "serde_json",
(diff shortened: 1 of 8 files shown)


kayagokalp (commented):
Should we also use it for:

https://github.com/FuelLabs/fuelup/blob/2d819944b17e14c633cf2769c88613133b8f7f0e/tests/default.rs#L60-L63

to remove the need of explicitly using `/n` in the expected case or if you think these are fine, happy to approve.

JoshuaBatty:
> Should we also use it for:
> 
> https://github.com/FuelLabs/fuelup/blob/2d819944b17e14c633cf2769c88613133b8f7f0e/tests/default.rs#L60-L63
> 
> to remove the need of explicitly using `/n` in the expected case or if you think these are fine, happy to approve.

Yeah good call, might as well add it here as well. Will push a commit now

JoshuaBatty:
actually i'll leave it as is for now as these contain bold formatting in places.

kayagokalp (approved):

PR #641, opened 2024-06-25
ci: enable test coverage tracking

kayagokalp:
Enables tracking of test coverage for PRs. Generates an `lcov` file with branch coverage, and uses an action to actually summarize the result to post comments to PRs. It is also possible to ensure a minimum coverage amount below which CI will fail but currently it is not being used.


sdankel (approved):
Nice one

Issue #652, opened 2024-08-15
When override toolchain from toml file is not installed, show a warning


sdankel:
There should be a warning message indicating that the toolchain from the override file needs to be installed. Currently it just shows "not found" for everything in `fuelup show`

alfiedotwtf:
This GitHub Issue currently only talks about `fuelup show` but I think all tools should be included in the fix... 

Here's the output for `fuelup show` on my machine:

```
> fuelup show
Default host: aarch64-apple-darwin
fuelup home: /Users/alfie/.fuelup

Installed toolchains
--------------------
beta-3-aarch64-apple-darwin (override)
latest-aarch64-apple-darwin (default)

active toolchain
----------------
beta-3-aarch64-apple-darwin (override), path: /Users/alfie/BACKUP/work/fuellabs/worktrees/counter-contract/master/fuel-toolchain.toml
  forc : 0.37.3
    - forc-client
      - forc-deploy : not found
      - forc-run : not found
    - forc-crypto : not found
    - forc-debug : not found
    - forc-doc : not found
    - forc-fmt : not found
    - forc-lsp : not found
    - forc-tx : not found
    - forc-wallet : 0.2.2
  fuel-core : 0.17.11
  fuel-core-keygen : not found

fuels versions
--------------
forc : 0.39
```

For me the confusing thing was when jumping straight into `forc build`:

```
> forc build
error: Found argument 'build' which wasn't expected, or isn't valid in this context

USAGE:
    forc [OPTIONS] <SUBCOMMAND>

For more information try --help
```

PR #656, opened 2024-08-26
Prompt user to install the override toolchain if its specified but not installed yet

alfiedotwtf:
This PR is in relation to #652

When overriding toolchains, there are 2 places to be concerned with when a toolchain is missing:

1. When `fuelup` is the command run
2. When a proxy command is run

For 2, we are already covering this scenario by installing missing components and plugins on first run.

As for 1, this PR will now `bail!()` unless the command is to `install`, `uninstall`, or create a `new` toolkit.

...

No tests yet because I want to get feedback to see if I'm on the right track.

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -1,9 +1,9 @@
 use crate::{
     constants::{
-        CHANNEL_BETA_1_FILE_NAME, CHANNEL_BETA_2_FILE_NAME, CHANNEL_BETA_3_FILE_NAME,
-        CHANNEL_BETA_4_FILE_NAME, CHANNEL_BETA_5_FILE_NAME, CHANNEL_DEVNET_FILE_NAME,
-        CHANNEL_LATEST_FILE_NAME, CHANNEL_NIGHTLY_FILE_NAME, CHANNEL_TESTNET_FILE_NAME,
-        DATE_FORMAT_URL_FRIENDLY, FUELUP_GH_PAGES,
+        BETA_CHANNELS, CHANNEL_BETA_1_FILE_NAME, CHANNEL_BETA_2_FILE_NAME,
+        CHANNEL_BETA_3_FILE_NAME, CHANNEL_BETA_4_FILE_NAME, CHANNEL_BETA_5_FILE_NAME,
+        CHANNEL_DEVNET_FILE_NAME, CHANNEL_LATEST_FILE_NAME, CHANNEL_NIGHTLY_FILE_NAME,
+        CHANNEL_TESTNET_FILE_NAME, DATE_FORMAT_URL_FRIENDLY, GITHUB_USER_CONTENT_URL,
     },
     download::{download, DownloadCfg},
     toolchain::{DistToolchainDescription, DistToolchainName},
@@ -16,21 +16,6 @@ use std::{collections::BTreeMap, fmt::Debug};
 use time::Date;
 use toml_edit::de;
 use tracing::warn;
-
-pub const LATEST: &str = "latest";
-pub const STABLE: &str = "stable";
-pub const BETA_1: &str = "beta-1";
-pub const BETA_2: &str = "beta-2";
-pub const BETA_3: &str = "beta-3";
-pub const BETA_4: &str = "beta-4";
-pub const BETA_5: &str = "beta-5";
-pub const DEVNET: &str = "devnet";
-pub const TESTNET: &str = "testnet";
-pub const NIGHTLY: &str = "nightly";
-
-pub const CHANNELS: [&str; 9] = [
-    LATEST, NIGHTLY, BETA_1, BETA_2, BETA_3, BETA_4, BETA_5, DEVNET, TESTNET,
-];
 
 #[derive(Debug, Deserialize, Serialize)]
 pub struct HashedBinary {
@@ -68,7 +53,7 @@ fn format_nightly_url(date: &Date) -> Result<String> {
 }
 
 fn construct_channel_url(desc: &DistToolchainDescription) -> Result<String> {
-    let mut url = FUELUP_GH_PAGES.to_owned();
+    let mut url = format!("{}{}/gh-pages/", GITHUB_USER_CONTENT_URL, "fuelup");
     match desc.name {
         DistToolchainName::Latest => {
             if let Some(date) = desc.date {
@@ -136,7 +121,7 @@ If this component should be downloadable, try running `fuelup self update` and r
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{download::DownloadCfg, file::read_file};
+    use crate::{constants::CHANNELS, download::DownloadCfg, file::read_file};
 
     #[test]
     fn channel_from_toml() {
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -35,14 +35,11 @@ pub struct Package {
     pub fuels_version: Option<String>,
 }
 
-pub fn is_beta_toolchain(name: &str) -> bool {
-    name == BETA_1
-        || name == BETA_2
-        || name == BETA_3
-        || name == BETA_4
-        || name == BETA_5
-        || name == DEVNET
-        || name == TESTNET
+pub fn is_beta_toolchain(name: &str) -> Option<&str> {
+    BETA_CHANNELS
+        .iter()
+        .find(|&beta_channel| name.starts_with(beta_channel))
+        .copied()
 }
 
 fn format_nightly_url(date: &Date) -> Result<String> {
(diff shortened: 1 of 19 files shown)


fuel-service-user:
### [LCOV](https://github.com/marketplace/actions/report-lcov) of commit [<code>69176e8</code>](656/commits/69176e855cdab751684c015c82f7f1e12da3a4e7) during [CI #2038](../actions/runs/10669207000)
<pre>Summary coverage rate:
  lines......: 82.3% (2364 of 2873 lines)
  functions..: 44.6% (367 of 823 functions)
  branches...: 59.6% (272 of 456 branches)

alfiedotwtf:
Once I found that toolchain installations were not failing because of code, everything started falling into place...

I've left the `dialoguer` comment in `fmt.rs` so that you can all see how that library is not able to be tested without a real tty (which is a shame given it works really nicely). So as is, we are still currently prompting the user via the original hand-rolled `stdin` nibbler.

We should now be a bit more robust with this PR when parsing `<channel>-<date>-<target>`, however I have left out most of the `split on "-"` code which I believe is/was causing most of the problems when specifying channel/toolchains, because it may be breaking behaviour which I'll avoid for now.

But for something to keep in mind, in the future we're probably better off converting `split on "-"` to regexes.

alfiedotwtf:
I forgot to mention, with the toolchain spec here:

```
/// Parses a distributable toolchain description from a string.
///
/// The supported formats are:
///     <channel>
///     <channel>-<target>
///     <channel>-<YYYY-MM-DD>
///     <channel>-<YYYY-MM-DD>-<target>
///     <channel>-<target>-<YYYY-MM-DD>
/// The parsing begins from the end of the string, so the target is the last part of the string,
/// then the date and finally the name
impl FromStr for DistToolchainDescription {
```

I think `<channel>-<target>-<YYYY-MM-DD>` should be removed since it doesn't follow the `channel` format of `<channel>[-<date>]`.

--

Other quick change to the PR could be to move the URL concats into a separate function like `fn github_releases_download_url()`, but figured it was easy enough to leave it as is. Happy to change it if someone thinks it's better to go that way.

src/channel.rs
```diff
@@ -68,7 +50,7 @@ fn format_nightly_url(date: &Date) -> Result<String> {
 }
 
 fn construct_channel_url(desc: &DistToolchainDescription) -> Result<String> {
-    let mut url = FUELUP_GH_PAGES.to_owned();
+    let mut url = format!("{}{}/gh-pages/", GITHUB_USER_CONTENT_URL, "fuelup");
```
sdankel, line 53: I'd prefer to keep `FUELUP_GH_PAGES` as a constant and refactor this to not use a mutable url.

```
fn construct_channel_url(desc: &DistToolchainDescription) -> Result<String> {
    let channel_file = match desc.name {
        DistToolchainName::Latest => {
            if let Some(date) = desc.date {
                &format!("channels/latest/channel-fuel-latest-{date}.toml")
            } else {
                CHANNEL_LATEST_FILE_NAME
            }
        }

        DistToolchainName::Nightly => {
            let mut file = "".to_string();
            if let Some(date) = desc.date {
                file.push_str(&format_nightly_url(&date)?);
                file.push('/');
            }
            file.push_str(CHANNEL_NIGHTLY_FILE_NAME);
            &file
        }
        DistToolchainName::Beta1 => CHANNEL_BETA_1_FILE_NAME,
        DistToolchainName::Beta2 => CHANNEL_BETA_2_FILE_NAME,
        DistToolchainName::Beta3 => CHANNEL_BETA_3_FILE_NAME,
        DistToolchainName::Beta4 => CHANNEL_BETA_4_FILE_NAME,
        DistToolchainName::Beta5 => CHANNEL_BETA_5_FILE_NAME,
        DistToolchainName::Devnet => CHANNEL_DEVNET_FILE_NAME,
        DistToolchainName::Testnet => CHANNEL_TESTNET_FILE_NAME,
    };
    let mut url = format!("{}{}/gh-pages/", GITHUB_USER_CONTENT_URL, "fuelup");


    Ok(format!("{FUELUP_GH_PAGES}{channel_file}"))
}

```

src/channel.rs
```diff
@@ -50,14 +35,11 @@ pub struct Package {
...
-        || name == BETA_2
-        || name == BETA_3
-        || name == BETA_4
-        || name == BETA_5
-        || name == DEVNET
-        || name == TESTNET
+pub fn is_beta_toolchain(name: &str) -> Option<&str> {
+    BETA_CHANNELS
+        .iter()
+        .find(|&beta_channel| name.starts_with(beta_channel))
```
sdankel, line 41: Shouldn't this still be `name == beta_channel`?

src/commands/toolchain.rs
```diff
@@ -45,7 +45,7 @@ fn name_allowed(s: &str) -> Result<String> {
         None => s,
     };
 
-    if RESERVED_TOOLCHAIN_NAMES.contains(&name) {
+    if CHANNELS.contains(&name) {
```
sdankel, line 48: `RESERVED_TOOLCHAIN_NAMES` is different from `CHANNELS`. We want `stable` to be reserved even though it isn't currently a usable channel.

src/config.rs
```diff
@@ -25,7 +25,7 @@ impl Config {
                 .filter(|e| e.file_type().map(|f| f.is_dir()).unwrap_or(false))
             {
                 let toolchain = dir_entry.file_name().to_string_lossy().to_string();
-                if RESERVED_TOOLCHAIN_NAMES
+                if CHANNELS
```
sdankel, line 28: Same question here.

src/fmt.rs
```diff
@@ -57,4 +57,14 @@ pub fn ask_user_yes_no_question(question: &str) -> io::Result<bool> {
             return Ok(result);
         }
     }
+
+    // This is the dialoguer version of the prompter, but it's not testable
```
sdankel, line 61: Have you tried testing it with `rexpect`? Here's an [example](https://github.com/FuelLabs/sway/blob/a81c42eb18827c644f46a6119054d4dacb500f92/forc-plugins/forc-client/tests/deploy.rs#L1349-L1351)

src/fuelup_cli.rs
```diff
@@ -46,6 +51,41 @@ enum Commands {
 pub fn fuelup_cli() -> Result<()> {
     let cli = Cli::parse();
 
+    if let Some(toolchain_override) = ToolchainOverride::from_project_root() {
```
sdankel, line 54: nit: add a comment here explaining what this is checking

src/fuelup_cli.rs
```diff
@@ -46,6 +51,41 @@ enum Commands {
...
+        let toolchain = match DistToolchainDescription::from_str(&override_path) {
+            Ok(desc) => Toolchain::from_path(&desc.to_string()),
+            Err(_) => Toolchain::from_path(&override_path),
+        };
+
+        info!("Using override toolchain '{}'", &toolchain.name);
+
+        if !toolchain.exists() {
+            match cli.command {
+                Commands::Toolchain(_) => {
```
sdankel, line 65: We only want to do this for operations where the toolchain override will be used in some way, right? So we also don't need to do this for `fuelup update/upgrade/install` `fuelup completions` `fuelup check` `fuelup default`

We really only need it for proxy commands and for `fuelup component` which modifies the currently active toolchain.

For `fuelup show` it could be good to show the current state first and then prompt the user if they wish to install.

tests/default.rs
```diff
@@ -99,13 +99,50 @@ fn fuelup_default_nightly_and_nightly_date() -> Result<()> {
...
 
+#[test]
+fn fuelup_default_override_skip_install() -> Result<()> {
+    testcfg::setup(FuelupState::LatestWithBetaOverride, &|cfg| {
+        let output = cfg.fuelup_with_input(&["default"], b"N\n");
+        let triple = TargetTriple::from_host().unwrap();
+
+        assert!(output
+            .stdout
+            .starts_with(&format!("Using override toolchain 'beta-1-{triple}'")));
```
sdankel, line 110: nit: we're planning to remove support for betas 1-4 pretty soon, so it might be better to make these tests use beta-5

tests/testcfg/mod.rs
```diff
@@ -163,6 +163,47 @@ impl TestCfg {
     pub fn fuelup(&mut self, args: &[&str]) -> TestOutput {
         self.exec("fuelup", args)
     }
+
+    /// A convenience wrapper for executing 'fuelup' within the fuelup test configuration.
+    /// It takes an array ref of bytes to write into stdin.
+    pub fn fuelup_with_input(&mut self, args: &[&str], input: &[u8]) -> TestOutput {
```
sdankel, line 169: This should be easier with `rexpect`

sdankel (changes requested):
It's coming along! I left some comments, and make sure to run `cargo fmt` to get CI passing.

alfiedotwtf:
_Keeping this PR in DRAFT but backgrounding for now..._

Most of these changes have moved to other PRs and have been merged, but there's a few change and comments remaining that I want to pull into future changes.

Issue #654, opened 2024-08-26
Possible inconsistencies with override versions


alfiedotwtf:
### Problem

Unless I'm mistaken, overrides works like the following:

> when an override file exists, then `forc`, **_all_** `components`, and `plugins` should use the overridden version rather than fall back to `default` unless an individual component or plugin is explicitly set within the override.

This doesn't seem to be the case:

```
> cat fuel-toolchain.toml 
[toolchain]
channel = "beta-3"

> forc --version
forc-tx 0.37.3

> forc-tx --version
component 'forc-tx' not found in currently active toolchain 'beta-3-aarch64-apple-darwin'

> fuelup component add forc-doc
Installing specific components is reserved for custom toolchains.
You are currently using 'latest-aarch64-apple-darwin'.

You may create a custom toolchain using 'fuelup toolchain new <toolchain>'.
```

### Steps

Let's start from scratch:

```
rm -f fuel-toolchain.toml



> fuelup show
Default host: aarch64-apple-darwin
fuelup home: /Users/alfie/.fuelup

Installed toolchains
--------------------
beta-3-aarch64-apple-darwin
latest-aarch64-apple-darwin (default)

active toolchain
----------------
latest-aarch64-apple-darwin (default)
  forc : 0.62.0
    - forc-client
      - forc-deploy : 0.62.0
      - forc-run : 0.62.0
    - forc-crypto : 0.62.0
    - forc-debug : 0.62.0
    - forc-doc : 0.62.0
    - forc-fmt : 0.62.0
    - forc-lsp : 0.62.0
    - forc-tx : 0.62.0
    - forc-wallet : 0.8.2
  fuel-core : 0.31.0
  fuel-core-keygen : 0.31.0

fuels versions
--------------
forc : 0.65.1
forc-wallet : 0.65.0



> fuelup component list
latest-aarch64-apple-darwin

Installed:
  forc 0.62.0 (latest: 0.63.1)
    - forc-fmt
    - forc-lsp
    - forc-doc
    - forc-deploy
    - forc-run
  forc-wallet 0.8.2 (latest: 0.9.0)
  fuel-core 0.31.0 (latest: 0.34.0)
  fuel-core-keygen 0.31.0 (latest: 0.34.0)

Installable:
```

Now we create the override:

```
> cat << EOF > fuel-toolchain.toml
[toolchain]
channel = "beta-3"
EOF



> cat fuel-toolchain.toml
[toolchain]
channel = "beta-3"



[...]

PR #665, opened 2024-10-04
Fix incorrect binary link target

alfiedotwtf:
When a proxy CLI'd command downloads a missing toolchain, multiple binary files within the downloaded distribution were being symlinked to the same one target file (thus overwriting each previous target).

This fixes an issue found while investigating #654.

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -385,16 +385,19 @@ impl Toolchain {
                 let store = Store::from_env()?;
                 for cfg in channel.build_download_configs() {
                     if store.has_component(&cfg.name, &cfg.version) {
-                        hard_or_symlink_file(
-                            &store
-                                .component_dir_path(&cfg.name, &cfg.version)
-                                .join(&cfg.name),
-                            &self.bin_path.join(&cfg.name),
-                        )?;
+                        self.add_component(cfg)?;
                     } else {
                         let downloaded = store.install_component(&cfg)?;
                         for bin in downloaded {
-                            hard_or_symlink_file(&bin, &self.bin_path.join(&cfg.name))?;
+                            // Use the actual binary filename rather than the
+                            // config name to prevent multiple binaries from
+                            // being linked to the same target file.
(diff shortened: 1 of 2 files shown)


fuel-service-user:
### [LCOV](https://github.com/marketplace/actions/report-lcov) of commit [<code>e500091</code>](665/commits/e500091940145fdba2f1a91cabec74ea40e72cdd) during [CI #2043](../actions/runs/11175211037)
<pre>Summary coverage rate:
  lines......: 81.8% (2315 of 2829 lines)
  functions..: 44.4% (363 of 818 functions)
  branches...: 58.4% (277 of 474 branches)

src/toolchain.rs
```diff
@@ -413,7 +413,12 @@ impl Toolchain {
                     } else {
                         let downloaded = store.install_component(&cfg)?;
                         for bin in downloaded {
-                            hard_or_symlink_file(&bin, &self.bin_path.join(&cfg.name))?;
+                            match bin.file_name() {
```
sdankel, line 416: If possible, it would be good to have a test that fails with the old code.

src/toolchain.rs
```diff
@@ -385,16 +385,16 @@ impl Toolchain {
...
-                                .component_dir_path(&cfg.name, &cfg.version)
-                                .join(&cfg.name),
-                            &self.bin_path.join(&cfg.name),
-                        )?;
+                        self.add_component(cfg)?;
                     } else {
                         let downloaded = store.install_component(&cfg)?;
                         for bin in downloaded {
-                            hard_or_symlink_file(&bin, &self.bin_path.join(&cfg.name))?;
+                            match bin.file_name() {
```
JoshuaBatty, line 392: I think some inline docs would be helpful .... 

feel free to tweak the below suggestion

```
// Use the actual binary filename rather than the config name to prevent
// multiple binaries from being linked to the same target file.
```

JoshuaBatty (approved):

PR #666, opened 2024-10-04
Extend finding components to plugins and executables distributed by forc

alfiedotwtf:
When checking if a component is distributed by `forc`, we currently only look at `forc`'s direct executables. This needs to be extended to resolve components, plugins, and executables.

This fixes a few issues found while investigating #665.

PR #668, opened 2024-10-04
Fix clobbering of process vs component vs forc

alfiedotwtf:
These issues were found while investigating #654. 

Currently, `component_name` was being repurposed for multiple functions:
- Checking whether or not the executable was published by `forc`
- Checking whether or not the executable was in the store
- Used to generate the tarball target
- Used to generate the tarball URL
- Used to generate the final target executable to run from the store

Each of the above was breaking in subtle ways depending on which executable was being run.

After separating out these functions based on the properly resolved component from the executable, and along with the previous commits, these problems above were resolved.

### DIFF
--- a/src/proxy_cli.rs
+++ b/src/proxy_cli.rs
@@ -6,7 +6,7 @@ use crate::{
     toolchain_override::ToolchainOverride,
 };
 use anyhow::Result;
-use component::Components;
+use component::{Component, Components};
 use std::{
     env,
     ffi::OsString,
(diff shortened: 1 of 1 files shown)


fuel-service-user:
### [LCOV](https://github.com/marketplace/actions/report-lcov) of commit [<code>e16411b</code>](668/commits/e16411bc2f6902d41bbfdf5250b3536f4a95e768) during [CI #2046](../actions/runs/11176479731)
<pre>Summary coverage rate:
  lines......: 82.3% (2374 of 2886 lines)
  functions..: 45.3% (375 of 827 functions)
  branches...: 59.8% (288 of 482 branches)

component/src/lib.rs
```diff
@@ -61,6 +74,24 @@ impl Component {
             && name != FORC)
             || name == FORC_CLIENT
     }
+
+    pub fn is_distributed_by_forc(component_name: &str) -> bool {
```
JoshuaBatty, line 78: Can we add doc comments for pub functions please?

component/src/lib.rs
```diff
@@ -61,6 +74,24 @@ impl Component {
...
+                .ok()
+                .and_then(|forc| {
+                    Component::resolve_from_name(component_name)
+                        .map(|component| Self::is_in_same_distribution(&forc, &component))
+                })
+                .unwrap_or(false),
+        }
+    }
+
+    pub fn is_in_same_distribution(forc: &Component, component: &Component) -> bool {
```
JoshuaBatty, line 91: Same here, doc comments..

JoshuaBatty: perhaps

```
/// Determines if two components are part of the same distribution
/// by comparing their repository names and tarball prefixes.
```

component/src/lib.rs
JoshuaBatty: wait looks like this is from #666 , can we add the comments to that PR?

JoshuaBatty:
Can we add some tests to show how this change works?

alfiedotwtf:
> Can we add some tests to show how this change works?

Yep, no problem.

alfiedotwtf:
Keeping this PR in DRAFT but backgrounding for now...

90% of the tests have been done, I just have to do the other 90%. Backgrounding for now, but will leave this as a downtime task to complete.

PR #667, opened 2024-10-04
Extend triple target to everything distributed by forc (#654)

alfiedotwtf:
Note: this is rebased on top of #666 👹

There are 2 types of downloads:
1. forc-binaries, which is of the format "{os}_{arch}"
2. Everything else, which is the standard "{arch}-{vendor}-{os}"

Currently, we only use the "{os}_{arch}" URL if the component is `forc`, however the `forc-binaries` tarball contains plugins and executables not defined in the `forc` component.

This PR extends the "{os}_{arch}" result to everything distributed from `forc`.

This fixes an issue found while investigating #654.

PR #683, opened 2024-10-24
Add tests and highlight issues with should_panic

alfiedotwtf:
This PR contains tests that demonstrate inconsistencies found in #654.

Once this PR is merged, the following PRs can then be updated along with removing corresponding `should_panic` test attributes:

- #665
- #666
- #667

Since #668 builds on top of the fixes in #666 and #667 and so will have conflicts and duplication, it will need to wait for the above PRs to be merged first before being mergeable.

### DIFF
--- a/component/src/lib.rs
+++ b/component/src/lib.rs
@@ -61,6 +61,11 @@ impl Component {
             && name != FORC)
             || name == FORC_CLIENT
     }
+
+    pub fn is_in_same_distribution(forc: &Component, component: &Component) -> bool {
+        component.repository_name == forc.repository_name
+            && component.tarball_prefix == forc.tarball_prefix
+    }
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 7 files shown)


fuel-service-user:
### [LCOV](https://github.com/marketplace/actions/report-lcov) of commit [<code>c4fa4da</code>](683/commits/c4fa4dadf9f556976d1b4977b792e5ea0fb8e65c) during [CI #2068](../actions/runs/11495541708)
<pre>Summary coverage rate:
  lines......: 86.3% (2475 of 2868 lines)
  functions..: 46.2% (381 of 824 functions)
  branches...: 63.0% (291 of 462 branches)

component/src/lib.rs
```diff
@@ -253,4 +258,131 @@ mod tests {
...
+        for executable in &Components::collect_plugin_executables().unwrap() {
+            let component = Component::from_name(executable).unwrap();
+            assert!(
+                component.executables.contains(executable),
+                "{} is an executable in {}",
+                executable,
+                component.name
+            );
+        }
+    }
```
alfiedotwtf, line 306: Although `Components::from_name(executable)` isn't something that we actually do now, it is used in that way in `direct_proxy()`. So I won't be modifying `from_name()` in the actual fix, but will follow #666 and add `resolve_from_name()` and use this instead where `Components::from_name(executable)` occurs

JoshuaBatty:
Nice work. Just wondering if we could look at adding some helper functions to cut down on code duplication. 

Something like

```rust
fn verify_toolchain_executables(cfg: &TestConfig, component: &Component, toolchain: &Toolchain) {
    let toolchain_bin_dir = cfg.toolchain_bin_dir(toolchain.name.as_str());
    for executable in &component.executables {
        assert!(
            toolchain_bin_dir.join(executable).exists(),
            "Executable not found: {}",
            executable
        );
    }
}
```

Probably a few places code duplication could be removed. Looks good otherwise.

component/src/lib.rs
```diff
@@ -61,6 +61,11 @@ impl Component {
             && name != FORC)
             || name == FORC_CLIENT
     }
+
+    pub fn is_in_same_distribution(forc: &Component, component: &Component) -> bool {
```
JoshuaBatty, line 65: perhaps a doc comment here like:

```
/// Checks if two components are part of the same distribution by comparing
/// their repository name and tarball prefix.
```

tests/testcfg/mod.rs
```diff
@@ -331,3 +384,31 @@ pub fn setup(state: FuelupState, f: &dyn Fn(&mut TestCfg)) -> Result<()> {
 
     Ok(())
 }
+
+pub fn verify_default_toolchain_override_toolchain_executables(
```
JoshuaBatty, line 388: can we add doc comments to pub functions

tests/testcfg/mod.rs
```diff
@@ -331,3 +384,31 @@ pub fn setup(state: FuelupState, f: &dyn Fn(&mut TestCfg)) -> Result<()> {
...
+    cfg: &TestCfg,
+    component: Option<&Component>,
+) {
+    let toolchain = get_default_toolchain_override_toolchain();
+    verify_toolchain_executables(cfg, component, &toolchain);
+}
+
+pub fn verify_toolchain_executables(
```
JoshuaBatty, line 396: same here

tests/testcfg/mod.rs
```diff
@@ -199,17 +212,66 @@ fn create_fuel_executable(path: &Path) -> std::io::Result<()> {
...
+//
+// ```no_run
+// use testcfg::{self, get_default_toolchain_override_toolchain, FuelupState};
+//
+// testcfg::setup(FuelupState::LatestToolchainInstalled, &|cfg| {
+//     let toolchain = get_default_toolchain_override_toolchain();
+//     assert_eq!(toolchain.name, format!("nightly-{}", yesterday()));
+// });
+// ```
+//
```
JoshuaBatty, line 274: not sure we need an empty line here and in the other examples....

JoshuaBatty (approved):

kayagokalp (approved):

kayagokalp:
It might be better to remove `(#654)` from titles of the PRs before merging otherwise they will show up as:

`Add tests and highlight issues with should_panic (#654) (#683)` in the history which might a little confusing as none of the other commits are in that form. Otherwise looks good to me!

PR #689, opened 2024-11-15
Add `forc-submit` to components.toml (sway #3875)

alfiedotwtf:
Sway's https://github.com/FuelLabs/sway/pull/3885 added a new executable to submit transactions to the network, but it was missed in the `fuelup` repo.

### DIFF
--- a/tests/show.rs
+++ b/tests/show.rs
@@ -36,6 +36,7 @@ fn fuelup_show_latest() -> Result<()> {
                 - forc-client
                   - forc-deploy : 0.1.0
                   - forc-run : 0.1.0
+                  - forc-submit : 0.1.0
                 - forc-crypto : 0.1.0
                 - forc-debug : 0.1.0
                 - forc-doc : 0.1.0
@@ -78,6 +79,7 @@ fn fuelup_show_and_switch() -> Result<()> {
                 - forc-client
                   - forc-deploy : 0.1.0
                   - forc-run : 0.1.0
+                  - forc-submit : 0.1.0
                 - forc-crypto : 0.1.0
                 - forc-debug : 0.1.0
                 - forc-doc : 0.1.0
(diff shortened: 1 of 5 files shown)


alfiedotwtf:
Weird...

Haven't dived in to find the reason behind this yet, but the tests are failing because Sway release binaries [1] contain `forc-submit` whereas the `sway-nightly-binaries` [2] don't 🤔

[1] https://github.com/FuelLabs/sway/releases/
[2] https://github.com/FuelLabs/sway-nightly-binaries/releases

alfiedotwtf:
With sway-nightly-binaries missing forc-submit [is now fixed](https://github.com/FuelLabs/sway-nightly-binaries/pull/55), I've run the build job to produce assets.

I'll update the `yesterday()` function to point to tomorrow so that this failing test will pick it up.

Then once tomorrow rolls over, all date hacks to `yesterday()` can then be removed.

fuel-service-user:
### [LCOV](https://github.com/marketplace/actions/report-lcov) of commit [<code>08adfca</code>](689/commits/08adfca4d742c7b04bcfd0601ae6779dea1b4495) during [CI #2096](../actions/runs/11869551644)
<pre>Summary coverage rate:
  lines......: 85.7% (2458 of 2868 lines)
  functions..: 46.1% (382 of 829 functions)
  branches...: 63.0% (290 of 460 branches)

alfiedotwtf:
@zees-dev Good catch. The second two, yes. But the first one, although I'm not actually sure where the line is to classify something to go into `forc` or a plugin, thinking it's the end of the pipe to `forc-tx` which is also not in `forc`, I'd leave it under `forc-client` for now.

(it actually doesn't make any difference AFAICS in the code).

tests/show.rs
```diff
@@ -36,6 +36,7 @@ fn fuelup_show_latest() -> Result<()> {
                 - forc-client
                   - forc-deploy : 0.1.0
                   - forc-run : 0.1.0
+                  - forc-submit : not found
```
sdankel, line 39: Shouldn't these show 0.1.0 as well?

alfiedotwtf: Doh! Oh wow, ok I see where I went wrong... I assumed that file didn't exist because it was pulling down an old version of from githubusercontent.com where `forc-submit` doesn't exist, but they're actually coming from mocked files defined by `ALL_BINS`. I'll fix that, thanks!

alfiedotwtf:
Thanks @sdankel, much better!

sdankel (approved):

PR #706, opened 2025-03-03
ci: update rust version to 1.85.0

sdankel:
Fixing CI errors seen in https://github.com/FuelLabs/fuelup/pull/705

Makes it so we:
- use the newest nightly for codecov workflow, which requires nightly (this fixes the codecov failure)
- uses a fixed version of rust for all of the other workflows (previously it was using the only nightly version for everything)
- this fixes the clippy error

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  RUST_VERSION: 1.84.1
+  RUST_VERSION: 1.85.0
   REGISTRY: ghcr.io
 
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  RUST_VERSION_COV: nightly-2024-06-05
+  RUST_VERSION: 1.84.1
```
zees-dev, line 13: current rust stable seems to be `1.85.0`; should we use the current latest stable here if possible?

PR #713, opened 2025-03-17
feat: fix install script mime type

mchristopher:
Adding a Cloudflare Worker which will re-write the `Content-Type` HTTP header with the correct MIME type for a shell script.

This is needed since GitHub pages doesn't support overriding the MIME type of a file, and the only "index" file that is supported must end in `.html` (which automatically gets the type `text/html`).


alfiedotwtf (approved):

sdankel (approved):
Looks good, thanks!

